PR #7, opened 2020-02-20
Generate types for API methods

azz:
👋 Hi @acacode, I have been searching the internets for a tool to convert an OpenAPI spec into TS declarations and I was surprised with the lack of quality in this area. None of the tools seemed to offer what I was after, which is simply Type Definitions (not even the generated client) for a REST API.

I then stumbled across your repo and found it super approachable to hack on and I got the result I was after in no time.

This PR adds exports for interfaces describing the `RequestQuery`, `RequestBody` and `ResponseBody` types for each route. These are very useful things to have as type annotations for function declarations dealing with API requests/responses.

Is this something you would like to see in the project?

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -202,9 +202,11 @@ const groupRoutes = routes => {
         )
         route.comments.push(`@originalName ${route.name}`)
         route.comments.push(`@duplicate`)
-        route.name += ++duplicates[route.moduleName][route.name];
+        const duplicateNumber = ++duplicates[route.moduleName][route.name]
+        route.name += duplicateNumber;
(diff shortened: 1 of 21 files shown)


js2me:
Hello @azz ! Thanks for your contribution. I'm glad to see new (not own) changes at this project! :)
This feature will really might be useful for people who use this tool.
I think this feature should be carried out by CLI option!
Also in future I want to separate all things (Api modules, Model interfaces) on CLI options

azz:
Thanks for the feedback, I'll fix that up shortly.

I was actually considering opening a second PR to add CLI flags, so I'm glad we're on the same page!

PR #15, opened 2020-02-22
Add --no-client and --route-types

azz:
This PR adds two CLI options:

- `--no-client` Don't generate the ApiClient class.
- `--no-route-types` Don't generate the per-route Body/Query/Response types.

I did this by splitting the mustache template into 3 files and rendering them conditionally.

I've tested it manually, but not sure how to go about adding tests for it?


index.js
```diff
@@ -16,15 +16,19 @@ program
   .description("Generate api via swagger scheme.\nSupports OA 3.0, 2.0, JSON, yaml.")
   .requiredOption('-p, --path <path>', 'path/url to swagger scheme')
   .option('-o, --output <output>', 'output path of typescript api file', './')
-  .option('-n, --name <name>', 'name of output typescript api file', 'api.ts');
+  .option('-n, --name <name>', 'name of output typescript api file', 'api.ts')
+  .option('--no-route-types', 'do not generate type definitions for API routes')
```
js2me: I think here should be reverse condition - `-r, --route-types`  
Also, please add default values (3 argument)

azz: `commander` has special handling of `--no-*` options so `false` is the default.

js2me:
Good solution is to separate template on 3 parts 👍
I'll add common manual tests for this new flags
Thanks!

PR #17, opened 2020-02-25
[Features] - response declarations, "default" status codes as success response types

js2me:
- Response declarations in request rescription  
  - Types for bad responses  
  - `-d, --default-as-success` option. Allows to use "default" status codes as success response type


index.d.ts
```diff
@@ -10,6 +10,22 @@ interface BaseGenerateApiParams {
...
+  generateRouteTypes?: boolean;
+
+  /**
+   * do not generate an API class
+   */
+  generateClient?: boolean;
+
+  /**
+   * use "default" response status code as success response too.  
+   * Some swagger schemas have "default" response status code.
```
js2me: Probably needs to add more information here 😄

PR #60, opened 2020-06-25
Custom template

### LINKED ISSUE #54: Custom Templates
Hi! 
It'd be cool to be able to provide custom templates directory to this tool in order to tweak templates or just leverage the parser to generate custom models or something like that. It's all only about yet another argument e.g. `--templates`.

Would it be of your interest if I made a PR with this feature?

andrefilimono:
This feature (closes #54) enables users to generate models and a client by providing path to a directory containing custom templates using a new CLI parameter `-t --template`.
It's also allows us to skip unnecessary templates if generating of a client or route types is not needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,43 +6,36 @@
 // License text available at https://opensource.org/licenses/MIT
 // Repository https://github.com/acacode/swagger-typescript-api
 
-const program = require('commander');
-const { resolve } = require('path');
-const { generateApi } = require('./src');
-const { version } = require('./package.json');
+const program = require("commander");
+const { resolve } = require("path");
+const { generateApi } = require("./src");
+const { version } = require("./package.json");
 
 program
-  .version(version, '-v, --version', 'output the current version')
+  .version(version, "-v, --version", "output the current version")
   .description("Generate api via swagger scheme.\nSupports OA 3.0, 2.0, JSON, yaml.")
-  .requiredOption('-p, --path <path>', 'path/url to swagger scheme')
-  .option('-o, --output <output>', 'output path of typescript api file', './')
-  .option('-n, --name <name>', 'name of output typescript api file', 'Api.ts')
+  .requiredOption("-p, --path <path>", "path/url to swagger scheme")
+  .option("-o, --output <output>", "output path of typescript api file", "./")
+  .option("-n, --name <name>", "name of output typescript api file", "Api.ts")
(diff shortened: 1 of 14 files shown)


js2me:
@andrefilimono can you do rebase from the `next` branch and `push --force-with-lease` to remove my commits from history?

index.js
```diff
@@ -52,5 +45,6 @@ generateApi({
   defaultResponseAsSuccess: defaultAsSuccess,
   generateResponses: responses,
   input: resolve(process.cwd(), path),
-  output: resolve(process.cwd(), output || '.')
-})
\ No newline at end of file
+  output: resolve(process.cwd(), output || "."),
+  templates: resolve(process.cwd(), templates),
```
js2me, line 49: `process.cwd()` returns path where this script has been executed and the default value of `templates` option will resolve wrong path for internal(default) templates.   
I think we need to remove default value from commander option and change that line to:  
```js
templates: resolve(templates ? process.cwd() : __dirname, templates || "./src/templates"),
```

src/templates.js
```diff
@@ -0,0 +1,20 @@
...
+
+  return {
+    apiTemplate: getTemplate("api"),
+    clientTemplate: config.generateClient ? getTemplate("client") : null,
+    routeTypesTemplate: config.generateRouteTypes ? getTemplate("route-types") : null,
+  };
+};
+
+const getTemplate = (templateName) =>
+  getFileContent(resolve(__dirname, `${config.templates}/${templateName}.mustache`));
```
js2me, line 16: Here is we have the same problem. `config.templates` already have absolute path to our templates folder and because of this we should change that line something like:   
```js
const getTemplate = (templateName) =>
  getFileContent(resolve(config.templates, `./${templateName}.mustache`));
```

src/index.js
```diff
@@ -34,7 +35,7 @@ module.exports = {
     output,
     url,
     name,
-    templates = config.templates,
+    templates = resolve(__dirname, config.templates),
```
andrefilimono, line 38: We need this for the case when the tool is called from another script and the `templates` argument is not passed, so a default value defined in the config should be resolved to internal templates.

There's might be a better and cleaner solution, but I unfortunately haven't found it, what do you think?

js2me: That's a right and a good solution!

js2me:
@andrefilimono Thank you for this PR with a nice feature!

PR #84, opened 2020-10-16
fix: security configuration in methods, When the security definition is in the main configuration of the swagger definition

jomarquez21:


js2me:
Thank you for this bugfix, I'll update swagger-typescript-api as soon as possible

PR #97, opened 2020-12-18
Generation of the module name: added the ability to pass the url inde…

nikalun:
On my project, when generating api using your library, we faced the problem that all urls of requests for api start with / api / and we generate one large module with all methods inside the api. In the code, I saw that the selectable index was hardcoded. Perhaps it will be convenient not only for me if it will be possible to select the index through the config.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,6 +55,11 @@ interface GenerateApiParams {
    * generate js api module with declaration file (default: false)
    */
   toJS?: boolean;
+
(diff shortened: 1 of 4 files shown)


js2me:
Yes, got your problem, thanks for this PR with fix  
I will merge it today soon

nikalun:
Thank you. If I missed something and need to add, write, I'll do it.

PR #102, opened 2021-01-05
Bump axios due to vulnerability (also bumped lodash)

Mvbraathen:
Ran npm audit fix and god rid of axios vulnerability. I have run the tests and everything seems okay.


brookjordan:
How do we push this through?

PR #115, opened 2021-01-14
disable strict ssl

kel666:
Introducted boolean options to disable SSL strict


js2me:
@kel666 Thanks for this PR!

Issue #96, opened 2020-12-07
Request cancellation support


ApacheEx:
First, thanks for the awesome library.

Would be really cool if we can extend the default template to support request cancellation...
for example as it's described here: https://react-query.tanstack.com/docs/guides/query-cancellation#using-fetch

what do you think? (in fact it should be easy to change and I can contribute if you give this idea a green light)

p.s I know that there is a possibility to provider a custom template, but would be nice to support it out of the box.

js2me:
Hello @ApacheEx ! Thanks for sweet words :)      
I will think about it, I know the `request cancellation` it is very nice thing!   
But now I need to release `4.0.0` version with `ETA` templates, `--modular` option and other cool things, and after I'll add `request cancellation` support

emagnier:
Thanks for this nice library, awesome work!

Yes the request cancellation would be a nice addition. It's especially useful to avoid potential memory leaks when you build complex UI, so you can cancel a request if a UI component is unmount.

js2me:
Here is update http client template sketch with request cancellation support   
```ejs
<%
const { apiConfig, generateResponses } = it;
%>


type CancelToken = Symbol | string | number;

const abortControllersMap = new Map<CancelToken, AbortController>();

export type RequestParams = Omit<RequestInit, "body" | "method" | "signal"> & {
    secure?: boolean;
    cancelToken?: CancelToken;
}

export type RequestQueryParamsType = Record<string | number, any>;


interface ApiConfig<<%~ apiConfig.generic.map(g => g.name).join(', ') %>> {
    baseUrl?: string;
    baseApiParams?: Omit<RequestParams, "cancelToken">;
    securityWorker?: (securityData: SecurityDataType) => Omit<RequestParams, "cancelToken">;
}

<% if (generateResponses) {%>
/** Overrided Promise type. Needs for additional typings of `.catch` callback */
type TPromise<ResolveType, RejectType = any> = Omit<Promise<ResolveType>, "then" | "catch"> & {
    then<TResult1 = ResolveType, TResult2 = never>(onfulfilled?: ((value: ResolveType) => TResult1 | PromiseLike<TResult1>) | undefined | null, onrejected?: ((reason: RejectType) => TResult2 | PromiseLike<TResult2>) | undefined | null): TPromise<TResult1 | TResult2, RejectType>;
[...]

Issue #14, opened 2020-02-21
Default response body type.


js2me:
Probably it should be `void` type ? Now its `any`

azz:
`void` or `never`?e

js2me:
@azz `void` better because
> In contrast `never` is a type containing no values, which means that a function with this return type can never return normally at all. 

https://stackoverflow.com/questions/37910669/what-is-the-difference-between-never-and-void-in-typescript

PR #149, opened 2021-02-11
Release 5.0.0

js2me:
Fixes:  
- Request content types auto substitution  
  i.e. if request body is form data, then request body content type will be `multipart/form-data`  
- Strange method name (issue #152, thanks @RoXuS)  
- Hardcoded Content-Type causes issues with some endpoints (issue #153, thanks @po5i)  
- Critical bug with `:paramName` path params (issue #154)  

Features:  
- Ability to provide custom formatting `fetch` response  
- `"IMAGE"` content kind for response\request data objects  
- `RequestParams` `RequestHeaders` types for `--route-types` (`routeTypes: true`) option (issue #150, thanks @Fabiencdp )  
- `--default-response` option. Allows to set default type for empty response schema (default: `void`) (based on issue #14)  
- Request cancellation support (issue #96, thanks @ApacheEx)  
  `RequestParams` type now have the `cancelToken` field  
  `HttpClient` instance now have the `abortRequest(cancelToken)` method  

BREAKING_CHANGES:  
- Fully refactored `http-client.eta` template, make it more flexible and simpler.  
  `HttpClient["request"]` takes one argument with type `FullRequestParams`  
  (previously it takes many count of arguments which was not flexible)  
- Changed the default response body type from `any` to `void` (issue #14)   

Internal:  
- Changed templates:  
  - `http-client.eta`  
  - `procedure-call.eta`  
  - `api.eta`  

This version works with previous templates.


templates/modular/http-client.eta
```diff
@@ -1,143 +1,207 @@
...
+<% if (generateResponses) {%>
 /** Overrided Promise type. Needs for additional typings of `.catch` callback */
-export type TPromise<ResolveType, RejectType = any> = Omit<Promise<ResolveType>, "then" | "catch"> & {
+type TPromise<ResolveType, RejectType = any> = Omit<Promise<ResolveType>, "then" | "catch"> & {
     then<TResult1 = ResolveType, TResult2 = never>(onfulfilled?: ((value: ResolveType) => TResult1 | PromiseLike<TResult1>) | undefined | null, onrejected?: ((reason: RejectType) => TResult2 | PromiseLike<TResult2>) | undefined | null): TPromise<TResult1 | TResult2, RejectType>;
     catch<TResult = never>(onrejected?: ((reason: RejectType) => TResult | PromiseLike<TResult>) | undefined | null): TPromise<ResolveType | TResult, RejectType>;
 }
 <% } %>
 
-export interface HttpResponse<D extends unknown, E extends unknown = unknown> extends Response {
```
Styn, line 23: This makes it hard to type responses in code that uses the HttpClient.
(For example when using this with the `call` method of `redux-saga` where you need to type the expected response yourself, since `call` uses `any`.

PR #168, opened 2021-02-18
Fix header overwrite in modular API template.

emilecantin:
If `requestParams` contained headers, the whole headers would be replaced by these ones, wiping out the automatic `Content-Type` header.

Putting the `requestParams` spread first fixes it, as the re-definition of `headers` will overwrite them.

### DIFF
--- a/templates/modular/http-client.eta
+++ b/templates/modular/http-client.eta
@@ -169,11 +169,11 @@ export class HttpClient<SecurityDataType = unknown> {
         return fetch(
         `${baseUrl || this.baseUrl || ""}${path}${queryString ? `?${queryString}` : ""}`,
         {
+            ...requestParams,
             headers: {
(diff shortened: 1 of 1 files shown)


js2me:
Hello @emilecantin !   
Thanks for PR with fix this problem!

templates/modular/http-client.eta
```diff
@@ -169,11 +169,11 @@ export class HttpClient<SecurityDataType = unknown> {
         return fetch(
         `${baseUrl || this.baseUrl || ""}${path}${queryString ? `?${queryString}` : ""}`,
         {
+            ...requestParams,
```
js2me, line 172: Can you add the same change for `default` `http-client.eta` template ?

js2me (approved):
LGTM! Thanks

Issue #176, opened 2021-02-19
The contentFormatter for ContentType:Json does not correctly format strings


Styn:
When sending a string as json, it should by json.stringified, otherwise the backend is not able to parse it correctly.

I'll create a PR for this

js2me:
Hello @Styn ! Thanks for your contribution !

js2me:
@Styn please, use the `next` branch!

Styn:
Oh sorry, I already created it, will try to fix

Styn:
My previous commits seem to also be included in that PR, even though I created a new fork...

js2me:
Fixed in 5.1.6

PR #177, opened 2021-02-19
Also Json stringify string data

Styn:
It might sound strange to stringify string data, but the alternative is to expect a user to wrap it in quotes him/herself 
```
`"${dataToPass}"`
```


Styn:
This fixes: https://github.com/acacode/swagger-typescript-api/issues/176

js2me:
Thank you @Styn for your contribution!

Issue #166, opened 2021-02-16
Allow overriding only one template


Nihisil:
Right now if you specify `--templates` option it makes the tool load all templates from the provided folder.

And if there are missing files in this folder it raises an error. 

I think it would be helpful to add fallback here and if in the provided folder template is not exists to load it from the default template folder.

js2me:
Hello @Nihisil , thanks for this issue!    
Yes, it is a good idea to customize only one template

js2me:
5.1.7 Release has partial fix for this issue (Allow to remove all unused templates except templates used in ETA `includeFile()`)

js2me:
It will fixed in 6.0.0 major version

js2me:
@Nihisil it should be fixed in 6.0.0  
Thank you for issue !

PR #179, opened 2021-02-22
Do not fail if template file does not exist

armsnyder:
This allows overriding specific template files and fallback on the defaults.

Before this change, the behavior was for the tool to fail with `Error: ENOENT: no such file or directory` if one of the template files in a custom templates directory does not exist.

Caveat: With this fix it will still error if the overridden template uses `includeFile` on a template file that is not overridden:

```
Error [Eta Error]: Could not find the template "./route-docs". Paths tried: /redacted/route-docs.eta
```

Example output:

```
✨ try to read templates from directory "/redacted"
❗❗❗ data contracts template not found in ./data-contracts.eta
Code generator will use the default template
❗❗❗ http client template not found in ./http-client.eta
Code generator will use the default template
❗❗❗ route types template not found in ./route-types.eta
Code generator will use the default template
❗❗❗ route name template not found in ./route-name.eta
Code generator will use the default template
☄️  start generating your typescript api
```

Related: #166

### DIFF
--- a/src/templates.js
+++ b/src/templates.js
@@ -1,6 +1,6 @@
 const _ = require("lodash");
 const Eta = require("eta");
-const { getFileContent } = require("./files");
+const { getFileContent, pathIsExist } = require("./files");
 const { resolve } = require("path");
 
 const getTemplates = ({ templates, modular }) => {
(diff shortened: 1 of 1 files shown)


js2me:
Hello @armsnyder !   
Thank you for your contribution!

PR #180, opened 2021-02-24
Release 5.1.7

js2me:
Fixes:  
- Do not fail if template file does not exist (issue #166, thanks @armsnyder )  
  Caveat: With this fix it will still error if the overridden template uses `includeFile` on a template file that is not overridden


js2me:
@all-contributors please add @armsnyder for code bug

PR #188, opened 2021-03-03
Split routes based on first tag

jnpoyser:
Adds an option to split the routes/API based on the first tag of the operation.

This feature is inspired by the `--module-name-index` option, which didn't quite work for our use case. We have more complex nested routes - like `/api/company/{companyId}/employee/{employeeId}`, which are driven by corporate security policies. Splitting on the index in the URL didn't work - in this example we have company operations and employee operations.

The solution: use the first tag for the operation as the module name. Tags in the openapi/swagger spec are collections/arrays, so should be stable.

For us, we then create a company class, an employee class, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Options:
                                 Also combine path params and query params into one object
   --module-name-index <number>  determines which path index should be used for routes separation (default: 0)
                                 (example: GET:/fruites/getFruit -> index:0 -> moduleName -> fruites)
+  --module-name-first-tag       splits routes based on the first tag
   --modular                     generate separated files for http client, data contracts, and routes (default: false)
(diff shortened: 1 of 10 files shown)


js2me:
Hello @jnpoyser !   
Can you run `test-all` script and refresh all generated apis from test schemas ?

js2me:
or you already did it ?

jnpoyser:
I did, but it generated a load of line break differences... didn't want to rely on it being cleared up on commit

Any tips?

js2me:
No, thank you for this feature! It really helpful! I'm sure because I did the same thing with using `hooks` in my job project!

js2me:
@all-contributors please add @jnpoyser  for code ideas

PR #193, opened 2021-03-04
Add missing union-enums type for node api

artsaban:


js2me:
@artsaban Thanks for this PR! Changes will be added in next release

PR #210, opened 2021-03-19
Fix boolean enum handling

haedaal:
Hello.
First, I really appreciate typescript-specialized codegen project!
I've got here searching for alternatives to swagger-codegen, openapi-generator as both of these didn't feel seamless & neat with typescript.

Trying swagger-typescript-api, I found this little bug. please give me a feedback :)

as-is:
```typescript
export type BooleanEnum = "true" | "false";
```

to-be:
```ts
export type BooleanEnum = true | false;
```


js2me:
Hello @haedaal !   
Thanks for this kind words!   
And for this good PR!

PR #214, opened 2021-03-27
Use "secure" option from global config

### LINKED ISSUE #212: "securityWorker" is only used if "secure" option is specified on each request
Hi! Thanks for the great package.

This seems to be a Fetch API variation of another issue - https://github.com/acacode/swagger-typescript-api/issues/192.

Here is what the `request` method looks like after generating the API with the following parameters:

```sh
$ yarn swagger-typescript-api -p ./source.json -o . -n result.ts
```

```ts
class Api {
  // ...

  public request = async <T = any, E = any>({
    body,
    secure,
    // ...
    ...params
  }: FullRequestParams): Promise<HttpResponse<T, E>> => {
    // the "secure" option from this.baseApiParams is not used here
    const secureParams =
      (secure &&
        this.securityWorker &&
        (await this.securityWorker(this.securityData))) ||
      {};

    const requestParams = this.mergeRequestParams(params, secureParams);

    // ...
  };
}
```

So, basically, when the Fetch API is used, there is no way to make use of `securityWorker` other than specifying `secure` on each request. With `axios` this is solved trivially with interceptors, but I'd like to continue using the Fetch API.

Here is a simple reproduction of the issue (v7.0.0, just run `$ yarn`): https://github.com/dkamyshov/swagger-typescript-api-bug-repro

| swagger-typescript-api | 6.4.2, 7.0.0 |
| ---------------------- | ------------ |

dkamyshov:
This PR fixes #212.

### DIFF
--- a/templates/base/http-clients/axios-http-client.eta
+++ b/templates/base/http-clients/axios-http-client.eta
@@ -69,7 +69,7 @@ export class HttpClient<SecurityDataType = unknown> {
         body,
         ...params
     }: FullRequestParams): Promise<AxiosResponse<T>> => {
-        const secureParams = (secure && this.securityWorker && (await this.securityWorker(this.securityData))) || {};
+        const secureParams = ((typeof secure === 'boolean' ? secure : this.baseApiParams.secure) && this.securityWorker && (await this.securityWorker(this.securityData))) || {};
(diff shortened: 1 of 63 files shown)


js2me:
@dkamyshov thank you for your contribution!

PR #218, opened 2021-04-01
Allow passing custom fetch function

sushruth:
This PR introduces a small feature where fetch function used within the HttpClient class can be customized.

Allowing to pass a custom fetch function enables a few things -

- Add implicit telemetry to the fetch function
- hook into before and after the function call is made in a generic way
- Handle special cases in a generic way around fetch function itself

Implements the feature from: #219


js2me:
Hello @sushruth !   
Thanks for your contribution!   
This changes will be added in the next release

Issue #222, opened 2021-04-04
Array query param values are serialized with the (non-default) comma separated style


Styn:
According to the OpenAPI spec the default serialization method of array query parameters is `style: form` and `explode: true` ([see docs here](https://swagger.io/docs/specification/serialization/))
Currently no different serialization methods are supported, so I think going with the default makes the most  sense.
This issue came to my attention since .Net Core by default doesn't understand the `explode: false` style, it expects the default.

PR #223, opened 2021-04-04
Fix: query param array serialization

Styn:
This is a fix for #222 
I also renamed the function to make that part of the code it bit clearer.


js2me:
Hello @Styn !  
Thank you for your fix!

PR #229, opened 2021-04-06
fix: incorrect handler for primitive parameters

prog13:


js2me:
Hello @prog13 !   
Thanks for your fix, It will be added 8.0.3

PR #232, opened 2021-04-08
disableProxy option, added catch to axios get

kel666:
Hi, recently I had problems using swagger-typescript-api at work.

After some investigations I found that our corporate proxy was the problem, and was simply solved by adding "proxy: false" to Axios options.

I also added a catch clause in the Axios GET to print out some infos and to avoid missing "catch" error from nodeJS.

Hope you will include it in next release!

Tnx
Fabio

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -77,6 +77,10 @@ interface GenerateApiParams {
    */
   disableStrictSSL?: boolean;
   /**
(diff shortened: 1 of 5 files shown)


js2me:
Hello @kel666 thanks for your PR!

Issue #236, opened 2021-04-12
[BUG] GET methos query option bug


rhkdgns95:
**Bug Report Quick Checklist**
- I am on the latest version of swagger-typescript-api (v8.0.3)
- I use package manager yarn.
- I run swagger-typescript-api on OS Window 10 PRO.
- I run swagger-typescript-api on Node.js v12.16.1

**Describe the bug**
```ts
(async () => {
  const res = await api.pet.findPetsByStatus({
    status: ["pending", "available"],
  });
  console.log("RES: ", res);
})();
```
![image](https://user-images.githubusercontent.com/46029124/114359750-3ecaeb00-9baf-11eb-8843-62d36e725732.png)

- It becomes a code generator normally.
- However, if you write an array string in the query content when requesting the GET method, the character value is requested along with the index value, not the string corresponding to the first index of the string array.

ex) https://petstore.swagger.io/v2/pet/findByStatus?status=available&status=pending&status=sold
![image](https://user-images.githubusercontent.com/46029124/114360020-88b3d100-9baf-11eb-8db4-dd55710f614f.png)

> Current situation
```
0=p&1=v     // pending [0] : "p" + available [1] : "v"
```

> Expected
```
status=pending&status=solid&status=available
```

**To Reproduce**
- \__generated__/index.ts
```ts
  private addArrayQueryParam(query: QueryParamsType, key: string) {
    const value = query[key];
    if (Array.isArray(value) && ["string", "number", "boolean"].includes(typeof value[0])) {
      return value.map((v) => `${key}=${v}`).join("&");
    } else {
      return `${value.map(this.addQueryParam).join("&")}`;
    }
  }
```

**Refs**
- https://petstore.swagger.io/

SaschaGalley:
I have created a pull request for this issue, hope this can get merged soon.

PR #237, opened 2021-04-13
fix: add array params with key

SaschaGalley:
this should fix the bug mentioned in #236 that was introduced in #229


js2me:
Thanks @SaschaGalley for your PR!

PR #238, opened 2021-04-13
Consider 2xx a successful status

wyozi:
OpenAPI allows specifying range response codes (https://swagger.io/docs/specification/describing-responses/). I don't think it's particularly good code style tbh, but "2xx" should still probably be considered a successful response


js2me:
Thanks @wyozi for PR!

PR #241, opened 2021-04-16
fix(types): add silent as param in declarations

mastermatt:
`silent` was added as an option in 6.1.0.
https://github.com/acacode/swagger-typescript-api/pull/185


js2me:
Hello @mastermatt, thanks for your PR!

Issue #243, opened 2021-04-21
is it possible axios file upload


guhyeon:
swagger is well,
and uploading file has been successed manually.
but it is failed with generated function.

generated function

```
  uploadPost = (data: { file?: File }, params: RequestParams = {}) =>
    this.request<UploadAttachFileResponse, void>({
      path: `/api/file/upload`,
      method: "POST",
      body: data,
      type: ContentType.FormData,
      ...params,
    });
```

manual successed code

```
 const form = new FormData();
    form.append('file', e.target.files[0]);
    const response = await axios.post(
      '/api/file/upload',
      form
    );
```

PR #244, opened 2021-04-21
axios file upload formData Type

guhyeon:
[is it possible axios file upload #243](https://github.com/acacode/swagger-typescript-api/issues/243)


js2me:
Hello @guhyeon , thanks for your contribution!

Issue #226, opened 2021-04-05
Axios Interceptors


indapublic:
Hello. Thanks for your great library.
Do you have good examples how to use axios interceptors (e.g. refresh token) for generated apis?

marko-hologram:
> 
> 
> Hello. Thanks for your great library.
> Do you have good examples how to use axios interceptors (e.g. refresh token) for generated apis?

I think you currently cannot add axios interceptors since you don't have access to the actual axios instance that is created by code generated by this library (at least I didn't manage to do it).

orouz:
not a great solution, but you can override the `instance` property on `HttpClient` to use your own.

i think it makes sense to pass an actual instance instead of an instance config.

indapublic:
@orouz I just saw recent release with `customFetch` feature 🤔

guhyeon:
1. change httpclient instance to not private 
![image](https://user-images.githubusercontent.com/9765614/115196381-ccb45200-a12a-11eb-8a2f-41e8cb3ebce0.png)

2.
` const api = new Api();
 api.instance.interceptors.response.use(...)
`

marko-hologram:
> 
> 
>     1. change httpclient instance to public
>        ![image](https://user-images.githubusercontent.com/9765614/115196381-ccb45200-a12a-11eb-8a2f-41e8cb3ebce0.png)
> 
> 
> `const api = new Api(); api.instance.interceptors.response.use(...)`

Wouldn't this change be reset back when you use `swagger-typescript-api` again since the file would be generated again? 
You would have to manually update it back to be public. Not that much of a hassle, but something to keep in mind.

guhyeon:
@marko-hologram, You are right. it is always reseted, unless you can use custom templates, successively.

I'm trying to use custom templates, but it is not work well.
so, I'm now change manually every generating. 🤔

guhyeon:
I just successed using template.

make directory "templates" and
make "http-client.eta" file in "templates" directory
and copy paste https://github.com/acacode/swagger-typescript-api/blob/next/templates/base/http-clients/axios-http-client.eta
and change private instance: AxiosInstance; => instance: AxiosInstance;

![image](https://user-images.githubusercontent.com/9765614/115210924-080a4d00-a13a-11eb-9d0e-124f7aa760fa.png)

marko-hologram:
> 
> 
> I just successed using template.
> 
> make directory "templates" and
> make "http-client.eta" file in "templates" directory
[...]

PR #245, opened 2021-04-23
Release 9.0.0

js2me:
NOTE: This version is not compatible with previous templates (removed `route.request.params`, `apiConfig.props`, `apiConfig.generic`, `apiConfig.description`, `apiConfig.hasDescription`)  

Fixes:  
- Consider 2xx a successful status (thanks @wyozi)  
- GET method query option bug (thanks @rhkdgns95, @SaschaGalley)  
- `silent` property missed in `.d.ts` file (thanks @mastermatt)  
- axios file upload `formData` type (thanks @guhyeon)  
- make property `instance` to public in axios http client (It can be helpful in #226)  
- variable name "params" doesn't uniq (thanks @mixalbl4-127 )  

Features:  
- `--disableProxy` option (thanks @kel666)  
- `--extract-request-body` option. Allows to extract request body type to data contract  
- Add TSDoc tag for deprecated route (thanks @wyozi)


indapublic:
```
--disableProxy option (thanks @kel666)
--extract-request-body option. Allows to extract request body type to data contract
```

Just wondering why naming so different. Probably `--disable-proxy` will be better for naming constancy

Issue #267, opened 2021-06-10
Generate Client without HttpResponse / ".data"


laktak:
Could swagger-typescript-api generate a client that returns
- `Promise<T>`
- instead of `Promise<HttpResponse<T, E>>`?

99% of the time I'm only interested in the actual data and doing `.data` makes the code less readable.

js2me:
It is only possible if you customize eta templates

laktak:
@js2me thanks, I added a possible solution.

rustyconover:
@js2me it might be good to document that using this makes the response headers inaccessible.

PR #268, opened 2021-06-14
added --unwrap-response-data to unwrap the data item from the response

laktak:
This is trying to solve #267.

Please let me know if this approach fits with your project.


js2me:
Hello @laktak !   
Thanks for this PR with the nice feature!   
Please make changes for axios-http-client eta template too

laktak:
@js2me I tried to add it according to the axios documentation.

I've tested the fetch implementation but I'm not using the axios client. Could you add some tests?

Liugq5713:
@laktak forget to change index.d.ts file ?

laktak:
> @laktak forget to modify index.d.ts file ?

Yes, I'm using the CLI version and did not notice that it had one.

PR #283, opened 2021-07-25
Feature: allow passing "patch" option to "swagger2openapi"

xcassel:
This PR adds the possibility to pass patch as an option.
This option is passed to the swagger2openapi converter to automatically **fix small errors in swagger definition code** like : 

- collectionFormat is only applicable to header.type array
- operation.consumes must be an array
- info.termsOfService must be a URL 
...

You can find all errors by searching (patchable) in [this file](https://github.com/Mermade/oas-kit/blob/706d2b8b548a0e13aabd5ff4811a6ba3784831a8/packages/swagger2openapi/index.js)

Without this option, the converter will throw an error and stop the generation.

Example : 

```
swagger-typescript-api\src\swagger.js:80
            throw new Error(err);
            ^

Error: S2OError: (Patchable) info.termsOfService must be a URL
    at swagger-typescript-api\src\swagger.js:80:19
    at swagger-typescript-api\node_modules\call-me-maybe\index.js:13:28
    at processTicksAndRejections (internal/process/task_queues.js:75:11)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Options:
   --default-response <type>     default type for empty response schema (default: "void")
   --type-prefix <string>        data contract name prefix (default: "")
   --type-suffix <string>        data contract name suffix (default: "")
+  --patch                       fix up small errors in the swagger source definition (default: false)
   -h, --help                    display help for command
 ```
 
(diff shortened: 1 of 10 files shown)


js2me (approved):

js2me:
@xcassel Thank you!

Issue #261, opened 2021-06-01
use .prettierrc when it exist (instead of default prettier config with tabWidth: 2)


mixalbl4-127:
@js2me https://prettier.io/docs/en/configuration.html

jjaros:
Hi, I simply updated the `scripts` in my project _package.json_ as follows:

```
  "scripts": {
    ...
    "generate-api": "ts-node build-scripts/generate-api && eslint --fix \"src/models/swagger-gen/*.ts?(x)\"",
    "build": "yarn generate-api && react-scripts build",
    ...
  },
```

`generate-api` - it starts the `swagger-typescript-api` (customized via _build-scripts/generate-api/index.ts_ file) and then starts the ESlint which also runs the `prettier`.  **The files `.eslintrc` and `.prettierrc` are used from the project.**

`build` - first generates the API files and then builds my React app in standard way

I think you can easily use this approach regardless of the framework (React, Angular, Vue, ...) or package manager (yarn, npm, ...) you use.

PR #286, opened 2021-08-14
Load prettier config from user's project

Argeento:
Search for prettier config based on [prettier docs](https://prettier.io/docs/en/configuration.html)

#261


js2me:
@Argeento thank you!

js2me (approved):

PR #294, opened 2021-09-20
Release 10.0.0

js2me:


hedgepigdaniel:
Merging this PR would be a good start? If this branch is already released to NPM, then why is the default branch `master` in a different version?

PR #295, opened 2021-09-22
fix(reject): fix reject and exit with code 1 if error

EvgenBabenko:
If we have an error in axios, for example invalid URL, we never catch this error in a Promise wrapper

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,29 +33,20 @@ program
   )
   .option(
     "-r, --responses",
-    "generate additional information about request responses\n" +
-      "also add typings for bad responses",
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -140,4 +140,9 @@ generateApi({
   typePrefix,
   typeSuffix,
   patch: !!patch,
+}).catch((err) => {
+  // NOTE collect all errors on top level and shows to users in any case
+  console.log(err);
```
js2me, line 145: I think it should be console.error, no ?

EvgenBabenko: my bad, fixed

js2me:
@EvgenBabenko Thank you!

PR #297, opened 2021-09-27
feat: add an option to disable throwing an error when !request.ok

RoCat:
In my app i have a lot of "normal" HTTP errors (e.g. 404 when no "object" found etc...).
It's a pain to alway surround API call with try / catch (or then / catch) to avoid code crash when a simple if(response.ok) will be enough.

I added "disableThrowOnError" option to remove the throw error behavior on fetch-http-client.
Default value is false, then the current behavior remain unchanged.


js2me:
Hello @RoCat ! Thanks for this feature request, can you switch target branch from master to next ?

RoCat:
Hi, change done ;)

js2me:
Thank you!

hognevevle:
Would love to see this option added to generateApi as well 🥳

arashi-dev:
It still isn't in the type definitions

PR #316, opened 2021-11-10
feat: add documentation and types for unwrapResponseData

simowe:
# Documentation and types for unwrapResponseData

I didn't know there was a unwrapResponseData feature until I randomly came across it in the templates.

This change makes the option easier to discover by adding it to README.md and enables it in TypeScript via index.d.ts


david-crespo:
I had the same experience. Thanks for adding this!

PR #325, opened 2021-12-13
document approach to disabling output

OutdatedVersion:
Looks like this is possible (not supported, maybe?) but not documented. Helps out a lot with using the exposed API with custom behavior on the tool's output.


js2me:
Hi @OutdatedVersion thanks for your contribution!

PR #332, opened 2022-01-12
Fix request format and allow for files in request

### LINKED ISSUE #251: Response `format` - global default or override ?
Hi

First thanks for great generator - really like it! Great work!

In release 8.0.0 there was [this commit](https://github.com/acacode/swagger-typescript-api/pull/225/commits/6427aa3463882c347d1382eb01d726d900eafa3d) which introduced feature "Allow to set global response type format through `HttpClient` constructor"

It was implemented both in Axios and Fetch clients as `const responseFormat = format && requestParams.format;` which works more like a global override (format passed into the constructor takes precedence over format passed into request)

It was later changed to `const responseFormat = format || requestParams.format;` with [this commit](https://github.com/acacode/swagger-typescript-api/pull/225/commits/6427aa3463882c347d1382eb01d726d900eafa3d) so it works more like a global default. But the **change was made only in Fetch template** - in Axios it is still [implemented](https://github.com/acacode/swagger-typescript-api/blob/0cbcbba33273d9ff309852e98b173821c8af6e44/templates/base/http-clients/axios-http-client.eta#L80) as global override...

As the current Fetch template behavior makes more sense, I guess it is actually a bug in the Axios template.

And that brings me to other part of the question. 
I do not want the Axios (or underlying `XmlHttpRequest`) to run a `JSON.parse` on my responses. Reason is I'm using custom [JSON.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) `reviver` function to parse all strings in the correct format directly to `Date`. So code generated with current version of `swagger-typescript-api` when used with current version of Axios (0.21.1) and passing `format: 'text'` into `Api` constructor works just fine. But if the problem mentioned above will be fixed and `format: 'json'` generated for my API routes starts overriding global default, it will break because `format` is passed into Axios param `responseType` and that is passed into `responseType` of `XmlHttpRequest` - as a result Axios starts to return JS objets parsed from JSON instead of just text...

This might not be an issue  in the end thanks to [this recent change in Axios](https://github.com/axios/axios/commit/5ad6994da3e97b8ed857faed90955a65a891f459) (not yet released) but I'm not sure (Axios development is a bit chaotic for me to follow) what is plan for the future (see [this change](https://github.com/axios/axios/pull/2424/files) in `1.0.0` branch)

So the question is, what is the easiest way to fix this ? Override the generation of `format: 'json'` for my routes ? Or something else ?

Thanks!

### LINKED ISSUE #293: Request falls if FormData attachment is File instance
Hi! I got issue with file format in FormData request. Library version: `9.0.2`.

Next entity is described in OpenApi format:

```yml
 FileSegmentCreate:
      type: object
      required:
        - files
      properties:
        files:
          type: array
          items:
            type: string
            format: binary
```

```ts
// Auto generated TS interface
export interface FileSegmentCreate {
    files: File[];
}
```

When I make request with file it falls because of the **incorrect file format**. After research I have found the trouble point in generated code, `contentFormatters` object. 

After `contentFormatters` code review I found out that:
1. it's not allowed to have multiply files in same FormData key. There is no check for array input type;
2. there is no check for Blob or File instance.

```ts
 // contentFormatters
 [ContentType.FormData]: (input: any) =>
      Object.keys(input || {}).reduce((data, key) => {
        data.append(key, input[key]);
        return data;
      }, new FormData()),
```

I'm able to resolve the instance issue on my side but It's hack.

```ts
//  createFileSegment expects that "files" key has File[] type;
const files = segment.files.map((file) => new Blob([file])) as any;
    
await api.createFileSegment({ ...segment, files: files[0] });
```

I noticed updates in `9.2.0` lib version that Blob instance check was added. However, the request will fall again if property is **File** instance or if it's Blob[] or Files[].

```ts
return Object.keys(input || {}).reduce((formData, key) => {
        const property = input[key];
        formData.append(
            key,
            property instanceof Blob ?
                property :
            typeof property === "object" && property !== null ?
                JSON.stringify(property) :
            `${property}`
        );
        return formData;
      }, new FormData())
```

## Questions

- is it able to add code to convert attachment to Blob if attachment has instance File?
- is it able to add check for detecting few attachments?

kuamanet:
Fix #293 
Fix #251

### DIFF
--- a/tests/spec/axiosSingleHttpClient/schema.ts
+++ b/tests/spec/axiosSingleHttpClient/schema.ts
@@ -1510,17 +1510,29 @@ export class HttpClient<SecurityDataType = unknown> {
     };
   }
 
+  private stringifyItem(item: any): string {
+    if (typeof item === "object" && item !== null) {
+      return JSON.stringify(item);
+    }
+
+    return `${item}`;
+  }
+
   private createFormData(input: Record<string, unknown>): FormData {
     return Object.keys(input || {}).reduce((formData, key) => {
       const property = input[key];
-      formData.append(
-        key,
-        property instanceof Blob
-          ? property
-          : typeof property === "object" && property !== null
-          ? JSON.stringify(property)
-          : `${property}`,
-      );
+      const propertyContent: Iterable<any> = property instanceof Array ? property : [property];
+
+      for (const formItem of propertyContent) {
+        const isFileType = formItem instanceof Blob || formItem instanceof File;
+        formData.append(key, isFileType ? formItem : this.stringifyItem(formItem));
+      }
+
+      return formData;
+    }, new FormData());
+    return Object.keys(input || {}).reduce((formData, key) => {
+      const property = input[key];
+      formData.append(key, property instanceof Blob ? property : this.stringifyItem(property));
       return formData;
     }, new FormData());
   }
@@ -1540,7 +1552,7 @@ export class HttpClient<SecurityDataType = unknown> {
         (await this.securityWorker(this.securityData))) ||
       {};
     const requestParams = this.mergeRequestParams(params, secureParams);
-    const responseFormat = (format && this.format) || void 0;
+    const responseFormat = format || this.format || undefined;
 
     if (type === ContentType.FormData && body && body !== null && typeof body === "object") {
       requestParams.headers.common = { Accept: "*/*" };
--- a/templates/base/http-clients/axios-http-client.eta
+++ b/templates/base/http-clients/axios-http-client.eta
@@ -66,16 +66,36 @@ export class HttpClient<SecurityDataType = unknown> {
         };
     }
 
+    private stringifyItem(item: any): string {
+      if (typeof item === "object" && item !== null) {
+        return JSON.stringify(item);
+      }
+
+      return `${item}`;
+    }
+
     private createFormData(input: Record<string, unknown>): FormData {
+      return Object.keys(input || {}).reduce((formData, key) => {
+        const property = input[key];
+        const propertyContent: Iterable<any> = (property instanceof Array) ? property : [property]
+
+        for (const formItem of propertyContent) {
+          const isFileType = formItem instanceof Blob || formItem instanceof File;
+          formData.append(
+            key,
+            isFileType
+              ? formItem
+              : this.stringifyItem(formItem),
+            );
+        }
+
+        return formData;
+      }, new FormData());
       return Object.keys(input || {}).reduce((formData, key) => {
         const property = input[key];
         formData.append(
             key,
-            property instanceof Blob ?
-                property :
-            typeof property === "object" && property !== null ?
-                JSON.stringify(property) :
-            `${property}`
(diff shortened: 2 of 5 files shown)


tests/spec/axiosSingleHttpClient/schema.ts
```diff
@@ -1511,6 +1511,23 @@ export class HttpClient<SecurityDataType = unknown> {
   }
 
   private createFormData(input: Record<string, unknown>): FormData {
+    return Object.keys(input || {}).reduce((formData, key) => {
+      let property = input[key];
```
NooNoo1337, line 1515: **suggestion:** lets make it as const. The value of variable doesn't change

tests/spec/axiosSingleHttpClient/schema.ts
```diff
@@ -1511,6 +1511,23 @@ export class HttpClient<SecurityDataType = unknown> {
...
+      let propertyContent: Iterable<any> = property instanceof Array ? property : [property];
+
+      for (const formItem of propertyContent) {
+        formData.append(
+          key,
+          formItem instanceof Blob || formItem instanceof File
```
NooNoo1337, line 1521: **suggestion:** to make external variable for condition to simplify double ternary condition.

```
const isFileType = formItem instanceof Blob || formItem instanceof File;
```

tests/spec/axiosSingleHttpClient/schema.ts
```diff
@@ -1540,7 +1557,7 @@ export class HttpClient<SecurityDataType = unknown> {
         (await this.securityWorker(this.securityData))) ||
       {};
     const requestParams = this.mergeRequestParams(params, secureParams);
-    const responseFormat = (format && this.format) || void 0;
+    const responseFormat = format || this.format || void 0;
```
NooNoo1337, line 1560: **suggestion:** void 0 -> undefined?

tests/spec/axiosSingleHttpClient/schema.ts
```diff
@@ -1511,6 +1511,23 @@ export class HttpClient<SecurityDataType = unknown> {
...
+    return Object.keys(input || {}).reduce((formData, key) => {
+      let property = input[key];
+      let propertyContent: Iterable<any> = property instanceof Array ? property : [property];
+
+      for (const formItem of propertyContent) {
+        formData.append(
+          key,
+          formItem instanceof Blob || formItem instanceof File
+            ? formItem
+            : typeof formItem === "object" && formItem !== null
```
NooNoo1337, line 1523: **suggestion:** get rid of the nested condition. Because it is hard to read and understand from scratch.

```
? formItem : stringifyFormItem(formItem)

stringifyFormItem(formItem) {
  if( typeof formItem === "object" && formItem !== null ) return JSON.stringify(formItem);
  
  return `${formItem}`;
}
```

kuamanet:
@NooNoo1337 change requests applied :)

NooNoo1337:
> @NooNoo1337 change requests applied :)

Awesome! Thank you c:

PR #335, opened 2022-01-17
Added "generateClient", "typePrefix", "typeSuffix" options to readme for NodeJS usage

Feargalicious:
Made changes as it took me a while to figure out the NodeJS option for --no-client is actually "generateClient" not "noClient"


js2me:
Hi @Feargalicious thanks for your contribution!

PR #336, opened 2022-01-19
Pass through required properties to allOf children

cyclops23:
When combining elements using `allOf` the required fields should be passed through so that the ts definitions do not define a required field in the child object as optional.  

I limited the scope of the changes to COMPLEX_ALL_OF as that is my use case, but happy to expand to other complex types for consistency if it makes sense.

### DIFF
--- a/src/schema.js
+++ b/src/schema.js
@@ -188,7 +188,7 @@ const complexSchemaParsers = {
   },
   [SCHEMA_TYPES.COMPLEX_ALL_OF]: (schema) => {
     // T1 & T2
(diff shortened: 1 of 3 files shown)


hedgepigdaniel:
Thanks for this, I had a similar problem!

I extended it to also work with `anyOf` and `oneOf`, but in the process I also found a bug in this change, where it can type fields as required when in fact they are not required.

See https://github.com/acacode/swagger-typescript-api/pull/342#discussion_r806545710 for an explanation.

cyclops23:
Closing in favour of #342

PR #342, opened 2022-02-15
Pass through required properties to allOf/anyOf/oneOf children

hedgepigdaniel:
Including https://github.com/acacode/swagger-typescript-api/pull/336, but:
 - extended to also work for `anyOf`/`oneOf`.
 - modified to avoid a bug where parent fields were incorrectly marked as required


tests/generated/v3.0/full-swagger-scheme.ts
```diff
@@ -15947,13 +15947,13 @@ export class Api<SecurityDataType extends unknown> extends HttpClient<SecurityDa
         | { status?: "queued" | "in_progress"; [key: string]: any }
         | ({ status?: "completed"; [key: string]: any } & { status?: "queued" | "in_progress"; [key: string]: any })
       ) & {
-        name: string;
+        name?: string;
```
hedgepigdaniel, line 15950: This is correct

tests/generated/v3.0/full-swagger-scheme.ts
```diff
@@ -15947,13 +15947,13 @@ export class Api<SecurityDataType extends unknown> extends HttpClient<SecurityDa
...
         head_sha: string;
         details_url?: string;
         external_id?: string;
         status?: "queued" | "in_progress" | "completed";
         started_at?: string;
-        conclusion?: "success" | "failure" | "neutral" | "cancelled" | "skipped" | "timed_out" | "action_required";
+        conclusion: "success" | "failure" | "neutral" | "cancelled" | "skipped" | "timed_out" | "action_required";
```
hedgepigdaniel, line 15956: This unfortunately is incorrect. This is the abbridged/relevant schema:
```
"schema": {
  "type": "object",
  "properties": {
    "name": {
      "type": "string",
      "description": "The name of the check. For example, \"code-coverage\"."
    },
    "status": {
      "type": "string",
      "description": "The current status. Can be one of `queued`, `in_progress`, or `completed`.",
      "enum": ["queued", "in_progress", "completed"]
    },
    "conclusion": {
      "type": "string",
      "description": "**Required if you provide `completed_at` or a `status` of `completed`**. The final conclusion of the check. Can be one of `success`, `failure`, `neutral`, `cancelled`, `skipped`, `timed_out`, or `action_required`.  \n**Note:** Providing `conclusion` will automatically set the `status` parameter to `completed`. Only GitHub can change a check run conclusion to `stale`.",
      "enum": [
        "success",
        "failure",
        "neutral",
        "cancelled",
        "skipped",
        "timed_out",
        "action_required"
      ]
    }
  },
  "anyOf": [
    {
      "properties": {
        "status": {
          "enum": ["completed"]
        }
      },
      "required": ["conclusion"],
      "additionalProperties": true
    },
    {
      "properties": {
        "status": {
          "enum": ["queued", "in_progress"]
        }
      },
      "additionalProperties": true
    }
  ]
},
```

src/schema.js
```diff
@@ -181,13 +181,19 @@ const filterContents = (contents, types) => _.filter(contents, (type) => !_.incl
 const complexSchemaParsers = {
   [SCHEMA_TYPES.COMPLEX_ONE_OF]: (schema) => {
     // T1 | T2
-    const combined = _.map(schema.oneOf, complexTypeGetter);
+    const combined = _.map(
+      schema.oneOf.map((s) => _.merge({ required: schema.required }, s)),
```
hedgepigdaniel, line 185: This logic isn't quite right, because it assumes that `schema.required` makes sense in the parent schema. But like the example in "full-swagger-scheme` shows, this isn't always the case.

Fixing this properly would require the logic in `complexSchemaParsers` to be lifted up a bit (i.e. if there is an `anyOf`/`oneOf`/`allOf`, then the entire type of the parent schema needs to be mapped, not just the fields in the `anyOf`.

But maybe a good first step is just to filter `schema.required` down to those fields that are directly defined in the nested `schema`. That way it would at least tighten the type for simple cases, without introducing this bug.

tests/generated/v3.0/full-swagger-scheme.ts
hedgepigdaniel: As in, `conclusion` is required, but only if `status` is `completed`. But the generated type indicates that it is required in all cases, which is incorrect.

hedgepigdaniel:
The last commit fixes the bug I mentioned with `required` referencing fields defined on the parent.

src/schema.js
```diff
@@ -178,16 +178,57 @@ const getObjectTypeContent = (schema) => {
...
+  // Identify properties that are required in the child schema, but
+  // defined only in the parent schema (TODO: this only works one level deep)
+  const parentPropertiesRequiredByChild = required.filter(
+    (key) => !_.keys(childSchema.properties).includes(key) && _.keys(parentSchema.properties).includes(key),
+  );
+
+  // Add such properties to the child so that they can be overriden and made required
+  properties = {
+    ...properties,
+    ...parentPropertiesRequiredByChild.reduce(
```
PatrickShaw, line 204: Why only fields required by the child?

To my knowledge `allOf` should apply ALL fields, regardless of whether they're optional/required. Are optional fields being handled elsewhere in this code base and if so shouldn't this code live in the same spot as that logic?

PatrickShaw: Maybe more importantly: Rather than overriding the original properties I believe the childSchema's properties and the parent schema's properties should be merged together so that both the child and parent's type conditions are validated against. As per: https://json-schema.org/understanding-json-schema/reference/combining.html#allof

> the given data must be valid against all of the given subschemas

Maybe it's just out of scope for this PR?

src/schema.js
```diff
@@ -178,16 +178,57 @@ const getObjectTypeContent = (schema) => {
 const complexTypeGetter = (schema) => getInlineParseContent(schema);
 const filterContents = (contents, types) => _.filter(contents, (type) => !_.includes(types, type));
 
+const makeAddRequiredToChildSchema = (parentSchema) => (childSchema) => {
+  let required = childSchema.required || [];
```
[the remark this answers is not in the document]
hedgepigdaniel: Yes, I tried to avoid that because personally (unfortunately) I still use this with Node.js 12. IMO the best way to address this is to add a build step, but that's out of scope for this PR.

src/schema.js
```diff
@@ -178,16 +178,57 @@ const getObjectTypeContent = (schema) => {
...
+
+  // Inherit all the required fields from the parent schema that are defined
+  // either on the parent schema or on the child schema
+  // TODO: any that are defined at grandparents or higher are ignored
+  required = required.concat(
+    (parentSchema.required || []).filter(
+      (key) =>
+        !required.includes(key) && (_.keys(properties).includes(key) || _.keys(parentSchema.properties).includes(key)),
+    ),
+  );
```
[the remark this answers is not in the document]
hedgepigdaniel: Yes, honestly I was just trying to stick to the convention, the surrounding code uses lodash data structures so I used them too :shrug: - happy to change if the maintainers prefer.

src/schema.js
```diff
@@ -178,16 +178,57 @@ const getObjectTypeContent = (schema) => {
...
+
+  // Identify properties that are required in the child schema, but
+  // defined only in the parent schema (TODO: this only works one level deep)
+  const parentPropertiesRequiredByChild = required.filter(
+    (key) => !_.keys(childSchema.properties).includes(key) && _.keys(parentSchema.properties).includes(key),
```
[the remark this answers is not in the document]
hedgepigdaniel: It could be, but I think `childSchema.properties` makes more sense here, since it's referring to the properties that are in the child schema, not the `properties` variable that will be modifiied and returned (which includes properties that are not in the child schema).

src/schema.js
hedgepigdaniel: > As per: https://json-schema.org/understanding-json-schema/reference/combining.html#allof

> > the given data must be valid against all of the given subschemas

I think that's referring to all of the **sub**schemas, not the combination of parent/children. Although yes, my understanding is that it must be valid against parent and child.

> Why only fields required by the child?
> the childSchema's properties and the parent schema's properties should be merged together

These fields aren't being added to check that their shape is correct. The generated types for oneOf/anyOf already inherit from the parent schema type at the typescript level - so effectively they are already merged.

The reason to add them here (duplicating the definitions from the parent) is so that they can be marked as required in typescript. Hence, this only includes fields that are defined in the parent, but made required in the child.

Another option might be to do it in typescript, e.g.
```typescript
type ChildSchema = ParentSchema & {
  child_field: string;
  parent_field_made_optional_in_the_child: Exclude<undefined, ParentSchema['parent_field_made_optional_in_the_child']>
};
```

But this requires the parent schema to have a name, which currently it doesn't.

src/schema.js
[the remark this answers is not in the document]
PatrickShaw: Yep fair enough. Leave it up to the author :)

tests/generated/v3.0/full-swagger-scheme.ts
```diff
@@ -10636,9 +10636,41 @@ export class Api<SecurityDataType extends unknown> extends HttpClient<SecurityDa
...
+              (
+                | { content: string }
+                | { filename: string | null }
+                | object
+                | ({ content: string } & { filename: string | null } & object)
+              ) & { content?: string; filename?: string | null }
+            >;
+          }
+        | ({ description: string } & {
+            files: Record<
```
hedgepigdaniel, line 10653: @PatrickShaw notice that `files` has exactly the same shape as the one on (new) line 10665. The entire definition is duplicated here - not because the shape is different, just so that it can be made a compulsory property. The type from the parent schema starts on line 10663 - and it has no name, so we can't refer to it here.

src/schema.js
hedgepigdaniel: > Are optional fields being handled elsewhere in this code base and if so shouldn't this code live in the same spot as that logic?

I'm not sure what you mean - open to suggestions if you have them.

src/schema.js
```diff
@@ -178,16 +178,57 @@ const getObjectTypeContent = (schema) => {
...
+  let required = childSchema.required || [];
+  let properties = childSchema.properties || {};
+
+  // Inherit all the required fields from the parent schema that are defined
+  // either on the parent schema or on the child schema
+  // TODO: any that are defined at grandparents or higher are ignored
+  required = required.concat(
+    (parentSchema.required || []).filter(
+      (key) =>
+        !required.includes(key) && (_.keys(properties).includes(key) || _.keys(parentSchema.properties).includes(key)),
```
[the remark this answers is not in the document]
hedgepigdaniel: I'm not sure, but given that logically `required` is a set, the most correct thing seems to be to avoid duplicates.

src/schema.js
```diff
@@ -178,16 +178,57 @@ const getObjectTypeContent = (schema) => {
 const complexTypeGetter = (schema) => getInlineParseContent(schema);
 const filterContents = (contents, types) => _.filter(contents, (type) => !_.includes(types, type));
 
+const makeAddRequiredToChildSchema = (parentSchema) => (childSchema) => {
```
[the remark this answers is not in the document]
hedgepigdaniel: Could be. I personally prefer creating a new variable instead of reassigning function parameters (destructured or not). But happy to change depending on what the maintainers prefer.

PatrickShaw (approved):

js2me:
Hello @hedgepigdaniel !  
Thanks for your PR!

js2me:
https://github.com/all-contributors please add @hedgepigdaniel for code bugs and add @PatrickShaw for review

js2me:
@all-contributors please add @hedgepigdaniel for code bugs and add @PatrickShaw for review

js2me:
@all-contributors please add @hedgepigdaniel for code bugs

js2me:
@all-contributors please add @PatrickShaw for review

Issue #304, opened 2021-10-09
Support for readOnly properties


kompasbank-niels:
May I suggest a small improvement. When swagger defines a property with "readOnly":true, then the generated property should also be marked as readonly.
This will help type-checking what is passed to an api, when the same type is used for input and response.

js2me:
Hello @kompasbank-niels !   
Yes, that will be a good idea, but this feature should work only with specific cli option like `--add-read-only`

PR #365, opened 2022-06-08
#304 Support readOnly properties

kmoskwiak:
Add readonly support #304


js2me:
@all-contributors please add @kmoskwiak  for code ideas

PR #420, opened 2022-10-19
feat(query-params): add an option to use brackets convention with arrays

qboot:
Hello :wave:

Currently the generated API encode query params arrays like: `?a=foo&a=bar&a=baz` with the `repeat` convention.
But all the backends don't follow this convention, and some use the `brackets` one: `?a[]=foo&a[]=bar&a[]=baz`.

So, this PR adds a new option `--query-params-with-brackets` which allow the use of the `brackets` convention instead of the `repeat` (only applicable when the query param is an array).


qboot:
@js2me can you elaborate on why this PR has been closed? Thx!

PR #422, opened 2022-10-19
chore: remove useless trailing whitespaces which make test edit harder

qboot:
If you are using VSCode and are trying to contribute to this project, there is a chance that you have the setting `"files.trimTrailingWhitespace": true,` enabled, like me.

And it took me some time to understand why my test suite was failiing.

I think this small fix can be useful even if you don't enforce any particular convention regarding trailing spaces.

What do you think?


qboot:
I've already run `npm test-all` :)

js2me:
@qboot thank you for your contribution !

PR #421, opened 2022-10-19
fix(deprecated): property with deprecated=false was shown as @deprecated

qboot:
We sometimes have in our OpenAPI schema some properties with the `"deprecated": false` flag.
These properties are annotated as `@deprecated` even with the flag set to `false`.

This PR fixes this issue and add a related test.

PR #428, opened 2022-10-24
feat(query-params): add an option to use brackets convention with arrays

qboot:
**NB: It's a repost of #420 which (I suppose) has been closed automatically!**

Hello :wave:

Currently the generated API encode query params arrays like: `?a=foo&a=bar&a=baz` with the `repeat` convention.
But all the backends don't follow this convention, and some use the `brackets` one: `?a[]=foo&a[]=bar&a[]=baz`.

So, this PR adds a new option `--query-params-with-brackets` which allow the use of the `brackets` convention instead of the `repeat` (only applicable when the query param is an array).

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Options:
   --another-array-type          generate array types as Array<Type> (by default Type[]) (default: false)
   --sort-types                  sort fields and types (default: false)
   --extract-enums               extract all enums from inline interface\type content to typescript enum construction (default: false)
+  --query-params-with-brackets use the `brackets` convention for array in query params: `?a[]=foo&a[]=bar` instead of `repeat` convention: `?a=foo&a=bar`
   -h, --help                    display help for command
 
 Commands:
--- a/index.d.ts
+++ b/index.d.ts
@@ -554,6 +554,7 @@ export interface GenerateApiConfiguration {
     cleanOutput: boolean;
     debug: boolean;
     anotherArrayType: boolean;
+    queryParamsWithBrackets: false;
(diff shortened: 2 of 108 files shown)


js2me:
Hello @qboot !  
Sorry for closing your PR without any comments.  
About your PR - this project is template driven and because of this all things which modifying inside templates is not needed because end user can modify templates as he\she\they want.   
One reason of modifying templates - enhancement core template functionality   
For example: providing more schema\other information for end user, improve additional jsdoc comments generation, ability to change typescript structs for interfaces\types\enums inside template (current it is not available)

js2me:
Also your previous PR was closed because it was target `master` which exists only for releases  
image

image

qboot:
Hello @js2me !

Thank you for your answer. Yes, we have managed to do it with templates in our project, but for me it was missing in the core code of the lib directly, as it's a pretty common thing to have to deal with different query params format.

And nope, I was targeting the `next` branch from the `master` branch of my fork (which was created from the `next` branch of the lib). Maybe it has confusef you.

And what about #421, it was a fix regarding an existant feature of the lib.

Thanks, @qboot

qboot:
Hello @js2me 
I have finally found the time to revamp this PR. For the record, it allows to use the `brackets` convention with the `fetch` API with an option. `axios` has an option to handle this case, but we need to handle it manually with `fetch`. 

I still find it relevant to do it internally in this lib instead of within my own templates because this case is really standard (it mimics `axios` + allows the lib to work with a wider variety of backends like Symfony for instance).

Thanks!

Issue #312, opened 2021-11-03
Content-Type text/plain is not respected


JochenDiekenbrock:
Given a path that accepts a put operation with a text/plain body:

```yml
info:
  title: Demo
  version: '1.0'
swagger: '2.0'
paths:
  '/agency':
    put:
      consumes:
        - text/plain
      produces:
        - text/plain
      parameters:
        - in: body
          name: agency
          required: true
          schema:
            type: string
            example: "123"
      responses:
        '200':
          description: OK
          schema:
            type: string
```

The generated client code for axios does not set the Content-Type. The Content-Type header in the http request is not set.
```typescript
agencyUpdate: (agency: string, params: RequestParams = {}) =>
      this.request<string, any>({
        path: `/agency`,
        method: "PUT",
        body: agency,
        ...params
      })
```

If in the generated code the type is added manually, the Content-Type header is set as specified:
```typescript
agencyUpdate: (agency: string, params: RequestParams = {}) =>
      this.request<string, any>({
        path: `/agency`,
        method: "PUT",
        type: "text/plain" as any,
        body: agency,
        ...params
      })
```

Expected behavior: The type should be set according to the "consumes" definition (and the typing of type should be adapted)

PR #443, opened 2022-11-04
Fix consumes test/plain

JochenDiekenbrock:
fixes acacode/swagger-typescript-api#312

The real changes are in

- index.d.ts
- src/schema-parser/schema-routes.js
- templates/default/procedure-call.ejs
- templates/modular/procedure-call.ejs
- tests/schemas/v2.0/api-with-examples.yaml
- templates/base/http-clients/axios-http-client.ejs
- templates/base/http-clients/fetch-http-client.ejs

The other changed files are the newly generated snapshots


js2me:
@JochenDiekenbrock thank you for your PR!

PR #476, opened 2022-12-25
fix: onCreateRoute skip behaviour

ivanovart:
I appears that the behaviour described in docs to skip the endpoint when `onCreateRoute` hook returns false was removed in v11
[change](https://github.com/acacode/swagger-typescript-api/commit/2d7925e96f87f269402122b00641652078f77d47#diff-481bff25593155cc0695aeb0f112c109e3075ef703d254734ec2a3c5266ee0beL676) -> `src/routes.js`


ivanovart:
@js2me, can you take a look please?

hanc2006:
I have the same issue. I cannot skip one route returning false

PR #486, opened 2023-01-24
Stop sending 'null' for null values in forms

### LINKED ISSUE #485: Null values in object being converted to form send "null" to server
When a request that uses the content type of `ContentType.FormData` is sent with a body, any null or undefined values in that object end up being sent to the server as `'null'` or `'undefined'` strings. This seems very wrong to make the server add special handling for strings containing the words null or undefined.

The code in question is this:
```ts
    [ContentType.FormData]: (input: any) =>
      Object.keys(input || {}).reduce((formData, key) => {
        const property = input[key];
        formData.append(
          key,
          property instanceof Blob
            ? property
            : typeof property === 'object' && property !== null
            ? JSON.stringify(property)
            : `${property}` // TODO: If property is null or undefined, this will result in 'null' or 'undefined'
        );
        return formData;
      }, new FormData()),
```

The line with the TODO shows the cause of the problem.
I see two ways of fixing this:
1. Wrap the `formData.append` call in a conditional to only add it to the FormData object if the value is not null or undefined.
2. Modify the line with the TODO so that it returns the empty string if `property` is null or undefined instead of blindly converting the value to a string.

To me, option 2 seems more correct since it's likely that purposefully setting a value to null or undefined may be a way of clearing a piece of data in the server; option 1 would not add the data to the form and thus wouldn't send the value to the server.

EdwardSalter:
When using a ContentType of FormData, requests will no longer send a string of `'null'` of `'undefined'` for any property in the body object that is null or defined.

The form value will now contain the empty string (`''`) instead.

Updated test snapshots to account for new template change.

The Fetch template now looks closer to the Axios template when it comes to checking for Forms and will also support File uploads, not just Blob uploads in the same manner as the Axios template.

Fixed a small bug in the package.json where the snapshots were not updating.

Fixes #485

### DIFF
--- a/templates/base/http-clients/fetch-http-client.ejs
+++ b/templates/base/http-clients/fetch-http-client.ejs
@@ -101,19 +101,26 @@ export class HttpClient<SecurityDataType = unknown> {
         return queryString ? `?${queryString}` : "";
     }
 
+    protected stringifyFormItem(formItem: unknown) {
+      if (formItem == null) {
+        return '';
+      } else if (typeof formItem === "object") {
+        return JSON.stringify(formItem);
+      } else {
+        return `${formItem}`;
+      }
+    }
+
     private contentFormatters: Record<ContentType, (input: any) => any> = {
         [ContentType.Json]: (input:any) => input !== null && (typeof input === "object" || typeof input === "string") ? JSON.stringify(input) : input,
         [ContentType.Text]: (input:any) => input !== null && typeof input !== "string" ? JSON.stringify(input) : input,
         [ContentType.FormData]: (input: any) =>
             Object.keys(input || {}).reduce((formData, key) => {
                 const property = input[key];
+                const isFileType = property instanceof Blob || property instanceof File;
                 formData.append(
                     key,
-                    property instanceof Blob ?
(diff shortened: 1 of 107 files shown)


templates/base/http-clients/fetch-http-client.ejs
```diff
@@ -101,19 +101,26 @@ export class HttpClient<SecurityDataType = unknown> {
         return queryString ? `?${queryString}` : "";
     }
 
+    protected stringifyFormItem(formItem: unknown) {
+      if (formItem == null) {
+        return '';
+      } else if (typeof formItem === "object") {
```
chiralium, line 107: There is useless `else-if` statements since you break the flow by `return`. You can simplify this part of code.

tests/generated/v2.0/another-schema.ts
```diff
@@ -124,21 +124,25 @@ export class HttpClient<SecurityDataType = unknown> {
     return queryString ? `?${queryString}` : "";
   }
 
+  protected stringifyFormItem(formItem: unknown) {
```
chiralium, line 127: DRY.

PR #487, opened 2023-01-26
Fix naming of disableStrictSSL to disableStrictSsl

max-zinn:
I recently upgraded from 9.3.1 to 12.0.2 and ran into the problem that the request to our swagger couldn't fulfill.

I found out it was because the `--disableStrictSSL` flag didn't work as intended so I debugged a little and could narrow the issue down to the fact that the `--disableStrictSSL` arrives as `disableStrictSsl` after processing the args.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -80,7 +80,7 @@ interface GenerateApiParamsBase {
   /**
    * disabled SSL check
    */
-  disableStrictSSL?: boolean;
+  disableStrictSsl?: boolean;
   /**
    * disabled Proxy
    */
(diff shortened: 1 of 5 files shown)


max-zinn:
Hi,
kindly asking, are there any updates on this?

@js2me

MadLord80:
Problem in file cli/process-option.js, function processFlags (camelCase):
```
if (!_.isEmpty(longestKey)) {
    name = _.camelCase(
      (isNoFlag ? longestKey.replace('--no-', '') : longestKey).replace(
        /(--?)/,
        '',
      ),
    );
  }
```

PR #495, opened 2023-02-21
Fix regex to match numbers in path parameters

koolmonkey:
Fix regex failure to match path parameters such as /api/store/{2dId}


js2me:
@koolmonkey thanks !!!

PR #498, opened 2023-02-28
Supporting exactOptionalPropertyTypes

seivan:
`signal` key has defined its key as optional, not necessarily the value even though this happens implicitly. No key means no value which translates to `undefined`:
```ts
    signal?: AbortSignal | null;
```

To support exactOptionalPropertyTypes it would have to be defined as:
```ts
    signal?: AbortSignal | null | undefined;
```

While you could say the key isn't optional (removing ?), which then requires an explicit undefined value:
```ts
    signal: AbortSignal | null | undefined;
```
That would require an explicit undefined being passed, which might break existing implementations, but also, it can't be changed here. 

Given this key is part of TS definitions and module augmentation is overkill, just "oring" from undefined to an expected value (in this case null) solves the problem.

Replacing: #457


js2me:
@seivan thanks!

PR #515, opened 2023-03-30
fix non-object custom spec extensions

### LINKED ISSUE #500: TypeError: Cannot create property 'consumes' on boolean 'true'
Hello Team

I am facing an error while generating typescript code with `x-internal: true` property. I checked open and closed issues but didn't find anything related to this hence creating an another issue. I am not sure if we already have any command line args to handle this. Please find the details below:

Screenshot from 2023-03-02 14-01-19

Here the code is broken:
Screenshot from 2023-03-02 14-03-59

Error message:
TypeError: Cannot create property 'consumes' on boolean 'true'

ajmnz:
Fixes #500


js2me:
@ajmnz thanks!

PR #538, opened 2023-06-05
feat: added hasSecurityData

Matzielab:
Added public hasSecurityData. This attribute can be used to check that securityData has been set before making a request or have an automatic hook call the api when securityData is set.


Matzielab:
Created this functionality using templates instead and doesnt seem to be any interest in having it by default so will close this PR

PR #541, opened 2023-06-15
Use type-only import for axios types

pataiadam:
### Refactoring the axios imports

The original code was importing both functions and types together from the axios library. 
To make the code more readable **and follow best practices**, I've separated the import statements into two lines. 

The first line now imports only the type definitions, i.e., AxiosInstance, AxiosRequestConfig, HeadersDefaults, and ResponseType from axios.
The second line imports the axios library itself.

Note:

_I've observed that in Vue projects built with Vite, the current approach **does not only** cause issues with the linter but **also leads to build failures**. The build process does not produce runnable code if type imports are not separated from library imports. Although the root cause of this behavior in Vite is unclear for me, these changes address this problem and ensure successful builds._

### DIFF
--- a/templates/base/http-clients/axios-http-client.ejs
+++ b/templates/base/http-clients/axios-http-client.ejs
@@ -2,7 +2,7 @@
 const { apiConfig, generateResponses, config } = it;
 %>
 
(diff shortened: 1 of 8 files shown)


js2me:
Thanks for pull request

PR #555, opened 2023-07-21
Update fetch-http-client.ejs to work correctly with FormData File[]

### LINKED ISSUE #556: Form Data "files" is not correctly handled with default fetch template
Form Data "files" is not correctly handled as an array of objects in fetch templates and is inconsistent with the axios templates. As a result, an array of files in FormData are sent in the fetch POST as an empty array.

omarmciver:
Update `fetch-http-client.ejs` to work correctly with FormData File[] and to be consistent with the same correct handling in the axios template.

Fixes #556

### DIFF
--- a/templates/base/http-clients/fetch-http-client.ejs
+++ b/templates/base/http-clients/fetch-http-client.ejs
@@ -104,19 +104,19 @@ export class HttpClient<SecurityDataType = unknown> {
     private contentFormatters: Record<ContentType, (input: any) => any> = {
         [ContentType.Json]: (input:any) => input !== null && (typeof input === "object" || typeof input === "string") ? JSON.stringify(input) : input,
         [ContentType.Text]: (input:any) => input !== null && typeof input !== "string" ? JSON.stringify(input) : input,
-        [ContentType.FormData]: (input: any) =>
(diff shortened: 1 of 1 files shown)


den9yon9 (approved):

templates/base/http-clients/fetch-http-client.ejs
```diff
@@ -104,19 +104,19 @@ export class HttpClient<SecurityDataType = unknown> {
...
-                    `${property}`
+        [ContentType.FormData]: (input) => Object.keys(input || {}).reduce((formData, key) => {
+            const property = input[key];
+            const propertyContent = (property instanceof Array) ? property : [property]
+
+            for (const formItem of propertyContent) {
+            const isFileType = formItem instanceof Blob || formItem instanceof File;
+            formData.append(
+                key,
+                isFileType ? formItem : JSON.stringify(formItem)
```
den9yon9, line 115: > isFileType ? formItem : typeof formItem === "object" && formItem !==null ? JSON.stringify(formItem): `${formItem}`

would be better

omarmciver: Code has been updated as suggested by @den9yon9 to cover the case where formItem is non-object or null.

PR #588, opened 2023-08-18
[CHG] Support for application/octet-stream

Shaenn:
Hello, 
Good job, for all, it's an amazing lib.

In my work, I use application/octet-stream for downloading file, but the parameter format wasn't set to 'blob'.

I can add a content-type image/png, but it's not really good, but it work...

I make this PR to resvolve this.

Best regards, Shaenn

### DIFF
--- a/src/schema-routes/schema-routes.js
+++ b/src/schema-routes/schema-routes.js
@@ -316,7 +316,8 @@ class SchemaRoutes {
     }
 
     if (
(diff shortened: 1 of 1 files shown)


planetdaz:
I could really use this too. My API Produces a content-type of application/pdf but the generated api request is setting to a format of json when it needs to be blob. 

Note: I'm using Axios, is there a way for me to override this format to get a requestType of 'blob'?

Thanks!

PR #600, opened 2023-09-01
Fix: FormData override

DeZZeDs:
Problem: If callback payload already instance of FormData then createFormData function his override value to empty
Fix: An additional condition has been added that checks for payload already instance FormData


yhostc:
A nice fix, thanks.

Issue #639, opened 2023-10-09
Option for handle object without $ref (nested objects)


frolovsky:
When generating the client, we pass the swagger. A swagger can include complex objects. Such objects usually have nested objects. So, if we want to access this nested object from code, then we can only get it if it is described inside the swagger using "$ref". 
Example object in schema (without $ref)
```json
"work_day": {
  "type": "object",
  "description": "Workday state",
  "properties": {
    "uuid": {
      "type": "string",
      "description": "Workday uuid",
      "example": "6e8242b9-adea-4d0a-ba74-07dc2cf6f459"
    }
  }
}
```

We generate some schema and result is:

```typescript
export interface Generated {
  name?: string;
  work_day?: {
    uuid?: string;
  };
}
```

But now, let use $ref inside our schema

```json
"work_day": {
  "schema": {
    "$ref": "#/components/schemas/Workday"
  }
}
```

Result in TypeScript:

```typescript

export interface Generated {
  name?: string;
  work_day?: GeneratedWorkDay;
}

export interface GeneratedWorkDay {
  uuid?: string;
}
```

It is possible to handle object without $ref?

frolovsky:
@js2me Hi! If I make changes that add an option for parsing, when passed, nested objects will be separated into a separate type, will you look at it? I see a lot opened pull request and as I understand it, there is not enough time to see everything)

PR #663, opened 2023-11-16
feat: add --extract-nested-objects option [issues/639]

frolovsky:
### Changes
New option `--extract-nested-objects` which parse interfaces recursivly. Related to https://github.com/acacode/swagger-typescript-api/issues/639

### Fixes
- add `extractResponses` to `index.d.ts` (for some reason, it wasn't there before)


dl-ias:
Waiting for this feature.

SmittyTint:
@smorimoto why was this branch closed? is there another way to handle this scenario?

frolovsky:
When I created the pull revext project, it had looked abandoned for several months. As I understand it, my pull request, like some others, was closed due to the need to refactor. Now these improvements are not relevant, but I plan to restore them. I also want to say thank you for the refactoring. 
P.S. Like for TS and for the vitest (tests take 23 seconds instead of 3 minutes, @smorimoto thx)

smorimoto:
@frolovsky We still have a lot of work to do, but thank you for your understanding 🙂

PR #631, opened 2023-09-27
fix: fix options manipulation

vgprst:
fixes #553 #585

PR #692, opened 2024-03-14
fix: define generateRouteTypes correctly

aleclarson:
The `--route-types` flag was being ignored, since the wrong option name from `program.execute` was being passed into the `generateApi` function.


JasonLandbridge:
@js2me Ping! This one is a feature fix, thanks @aleclarson!

aleclarson:
Duplicate of #631

PR #693, opened 2024-03-15
Consistent node-fetch usage (also fixes ECONNRESET error)

nicky1038:
There is an issue with the current version of `swagger-typescript-api`. After _successful_ generation of the file, indicated by this message:

`✅   typescript api file [FILE] created in [FOLDER]`

the execution does not stop. After some seconds of waiting I get this error in my console, and only then the utility stops.

```
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -4077,
  code: 'ECONNRESET',
  syscall: 'read'
}
```

I've discovered that the reason is `node-fetch-h2` package, just changing it to `node-fetch` solved the issue.

Moreover, right now there are some inconsistencies with using `node-fetch` and `node-fetch-h2` in the project:
* `node-fetch-h2` is used in the code but it is not stated in `package.json`
* `node-fetch` v3 is stated in `package.json`, it is used only for typings; v3 could not be used in this project as it uses ESM.

My solution is to use only `node-fetch`, the latest v2 version. With my fix its usage now looks as expected and also works good without any errors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "swagger-typescript-api",
-      "version": "13.0.2",
+      "version": "13.0.3",
       "license": "MIT",
       "dependencies": {
         "@types/swagger-schema-official": "2.0.22",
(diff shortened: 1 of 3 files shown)


smorimoto:
In #727, this fix is going to get covered, and node-fetch has been replaced with native fetch. Thanks a lot @nicky1038!

Issue #460, opened 2022-11-16
Query parameters with dots to camelCase


angela-patino-yuli:
**Summary**
A similar issue has been fixed for path parameters https://github.com/acacode/swagger-typescript-api/issues/413 
Is it possible to have it for query params as well? 
On my swagger.json file, If a query parameter has a dot in it (e.g. from.year) then swagger-typescript-api rewrites this to 
` "from.year"?: number;` instead of `"fromYear"`
    
Expected
 ```
foobar = (
    query?: {
      /** @format int32 */
      "fromYear"?: number;
    },
    params: RequestParams = {},
  ) =>
    this.http.request<void, any>({
      path: `/api/list`,
      method: "GET",
      query: query,
      secure: true,
      ...params,
    });
```

 Actual
```
 foobar = (
    query?: {
      /** @format int32 */
      "from.year"?: number;
    },
    params: RequestParams = {},
  ) =>
    this.http.request<void, any>({
      path: `/api/list`,
      method: "GET",
      query: query,
      secure: true,
      ...params,
    });
```

**Comment by js2me** [comment]
@angela-patino-yuli can you share part of your schema with this query params?

**Comment by js2me** [comment]
@angela-patino-yuli I think this can broke server side logic for other users, need more research to find a better way to do that.

**Comment by angela-patino-yuli** [comment]
thanks. if you need anything. let me know 😺

**Comment by adrivelasco** [comment]
@js2me hey, and what if I would like to keep dots?

**Comment by mr-raccoon-dev** [comment]
@adrivelasco I dunno did you find a solution or not, but you can use version 12.x.x to queries with dots.
Looks like it was broken since 13.0.0 version.
In v12.0.4 query parameters with dots works fine.

**Comment by depsimon** [comment]
@js2me it does indeed break server-side logic in my case at least.

The server expects dots in the parameter name.

In my case we have these parameters : 

```
{
    "name": "artist.id",
    "in": "query",
    "description": "",
    "required": false,
    "deprecated": false,
    "allowEmptyValue": true,
    "schema": { "type": "integer" },
    "style": "form",
    "explode": false,
    "allowReserved": false
},
{
    "name": "artist.id[]",
    "in": "query",
    "description": "",
    "required": false,
    "deprecated": false,
    "allowEmptyValue": true,
    "schema": { "type": "array", "items": { "type": "integer" } },
    "style": "form",
    "explode": true,
    "allowReserved": false
},
{
    "name": "id",
[...]

PR #701, opened 2024-04-19
fix: revert e340527 - removed forced renaming query params to camelCase

nicky1038:
There was a change introduced in the 13.0.0 major release. Since that release all query parameter names are being renamed into `camelCase`. **It is a way wrong behavior.**

Why is it wrong? Say, a backend server expects a query parameter called `a.b`. In the Swagger schema provided by that server, that query parameter is also called `a.b`. When `swagger-typescript-api` generates types, it renames query parameter to `aB` but does not provide any runtime transform. When user passes `aB` query parameter, the server obviously cannot recognize it, as it expects `a.b`.

I cannot understand what task is being solved with that change, as it is not valid at all and breaks logic.

It was suggested in #460. After it was released, numerous users also compained that their clients are generated wrongly.

My PR reverts that change.


smorimoto:
Could you rebase to latest main branch?

nicky1038:
@smorimoto Hi! Rebase done. Thank you so much for taking over maintenance of this project!

george-oakling:
Hello, this is a blocker for us to upgrade to v13 version.

harutlc:
this is blocker for me too

george-oakling:
Still no response?

MaximeDetaille:
I got the same issue ... please merge the PR

PR #750, opened 2024-05-28
feature: add --prefer-types-over-interfaces options

TiagoCerqueira:
Hello,

We have implemented this feature, with two reasons : 

- We prefer use types over interfaces for our domain
- In our project we have a french named domain named "Document" and the name clash with the "Document" from MDN. https://developer.mozilla.org/en-US/docs/Web/API/Document


kortirso:
really required feature

Issue #780, opened 2024-06-19
Code formatting broken since 13.0.5


pebo:
The https://github.com/acacode/swagger-typescript-api/pull/717 PR breaks code formatting if the prettier config file is not located in the current working directory.

From [v9.0.0](https://github.com/cosmiconfig/cosmiconfig/releases/tag/v9.0.0) the cosmiconfig package no longer searches upwards for prettier config by default. To enable the old behavior the `searchStrategy` `global` should be used.

NiiMER:
Any update on this from The @ Contributors ??

NiiMER:
Would it be ready in the next version @smorimoto ?

smorimoto:
@pebo @NiiMER https://github.com/acacode/swagger-typescript-api/releases/tag/13.0.7

PR #781, opened 2024-06-19
fix: support shared prettier config files

pebo:
Revert breaking change introduced in cosmiconfig v9.0.0

Fixes https://github.com/acacode/swagger-typescript-api/issues/780


smorimoto (approved):

NiiMER (commented):
LGTM

PR #742, opened 2024-05-19
fix: format option of requests when using extractResponseBody

TaopaiC:
When using the `--extract-response-body` option, the format option of requests disappears.

For example (from `tests/spec/extractResponseBody.js`):

When running with extractResponseBody: true, the generated code looks like this:

```typescript
this.request<FindPetsByStatusDataTTT, void>({
  path: `/pet/findByStatus`,
  method: "GET",
  query: query,
  secure: true,
  ...params,
}),
```

However, when running without `extractResponseBody: true`, the generated code includes the format option:

```typescript
this.request<PetTTT[], void>({
  path: `/pet/findByStatus`,
  method: "GET",
  query: query,
  secure: true,
  format: "json",   // <- this
  ...params,
}),
```

I attempted to copy `contentKind` to make the format is in the generated request options. I'm not sure if this is the good approach.

PR #826, opened 2024-07-11
fixed missing format when the response schemas are extracted to types

raing3:
This may not be the most appropriate way to fix the problem but hopefully the change/details below make it clear enough what issue I'm trying to address.

The issue is:

1. If extractResponseBody is set to true in the params the contentKind on the schema is being "lost".
2. That results in the format not being defined in the object constructed when sending a API call.
3. Because the format is missing the client will not attempt to decode the JSON data in the [http client](https://github.com/acacode/swagger-typescript-api/blob/main/templates/base/http-clients/fetch-http-client.ejs#L196).

This behaviour differs to when extractResponseBody is false where format would be populated based on the API spec. IMO this is unexpected. It doesn't feel appropriate for this behaviour to differ based on the inline response schemas are extracted to named types or not.

... thanks for the prompt merge on my earlier MR as well. :)

### DIFF
--- a/src/schema-routes/schema-routes.js
+++ b/src/schema-routes/schema-routes.js
@@ -729,12 +729,14 @@ class SchemaRoutes {
       let successResponse = responseBodyInfo.success;
 
       if (successResponse.schema && !successResponse.schema.$ref) {
+        const contentKind = successResponse.schema.contentKind;
         const schema = this.getSchemaFromRequestType(successResponse.schema);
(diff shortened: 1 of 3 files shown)


smorimoto:
It seems like the duplicate PR has already been opened: https://github.com/acacode/swagger-typescript-api/pull/742

raing3:
Ah, missed it, yep. Thanks for pointing it out.

PR #828, opened 2024-07-15
Custom config: handle default export

### LINKED ISSUE #806: Version 13.0.10 doesn't respect `onInsertPathParam` hook
I'm using the following custom configuration:
```javascript
module.exports = {
  hooks: {
    onInsertPathParam: (paramName) => `encodeURIComponent(${paramName})`,
  },
};
```
This encodes all parameters using `encodeURIComponent` in the URL path. This works great with v13.0.9, but it doesn't work with v13.0.10 anymore.

This is generated with v13.0.9:
```typescript
    deleteBucket: (name: string, params: RequestParams = {}) =>
      this.request<void, ApiError>({
        path: `/buckets/${encodeURIComponent(name)}`,
        method: "DELETE",
        secure: true,
        ...params,
      }),
```

This is generated with v13.0.10:
```typescript
    deleteBucket: (name: string, params: RequestParams = {}) =>
      this.request<void, ApiError>({
        path: `/buckets/${name}`,
        method: "DELETE",
        secure: true,
        ...params,
      }),
```

shrpne:
It should finally fix #806

Added interop default, because currently importing of custom config with `export default` is not working properly.

Also reorganized custom config logs a bit:
- logging "found custom config" only if it really loads
- otherwise showing error, instead of silently ignoring custom config

### DIFF
--- a/index.js
+++ b/index.js
@@ -296,14 +296,18 @@ const main = async () => {
   const { command, options } = await program.execute({ args: process.argv });
 
   let customConfig = null;
+  let customConfigPath;
 
   if (options.customConfig) {
     try {
-      const customConfigPath = resolve(process.cwd(), options.customConfig);
-      console.log(`✨ found custom config at: ${customConfigPath}`);
+      customConfigPath = resolve(process.cwd(), options.customConfig);
       customConfig = await import(customConfigPath);
+      customConfig = customConfig.default || customConfig;
     } catch (e) {
-      /* empty */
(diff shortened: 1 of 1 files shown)


smorimoto (approved):

smorimoto:
Thanks!

xAl3xFx:
In order for this to work with windows you need to add "file://" before `process.cwd()`.

PR #849, opened 2024-07-25
feat: add extra-templates modular approach

vkefallinos:


Mahdhir:
Coincidentally was working on the same thing 4 hours ago 😆

vkefallinos:
Nice. I also want to add the ability to handle the generated file naming on the default templates.

Mahdhir:
> Nice. I also want to add the ability to handle the generated file naming on the default templates.

You might want to take a look at hooks. Seems like you can handle your scenarios in that.

Issue #834, opened 2024-07-17
Breaking change in node 18 version requirement in #727


sparr:
#727 added a dependency on `"engines": { "node": ">=18.0.0" }` which caused build failures in my node 16 project that is using this package. That change should have required at least a patch version bump so that a `~13.0.3` dependency would not have updated to an incompatible version, and there's an argument to be made that it should have been a major version bump so that `^13.0.3` would also remain unbroken.

smorimoto:
Engines is a constraint that can be ignored, and it is a recommendation rather than a constraint. Simply add a setting to ignore and stop using the EOL version.

PR #851, opened 2024-07-28
Prepare 13.0.14

smorimoto:


sparr:
I'd like to call attention to #834 and suggest that this version should be 13.1.0 or 14.0.0 if the node version dependency change isn't rolled back.

PR #859, opened 2024-08-01
Make types slightly better

smorimoto:


smorimoto:
I don't actually use these types, so this may not meet your needs depending on your usage, but we have to go ahead anyway 🤷‍♂️

PR #747, opened 2024-05-24
Fix contentFormatter for FormData returning an empty FormData object

JorensM:
This PR fixes an issue where `contentFormatter` was returning an empty `FormData` object, effectively not allowing to submit any form data when the content type is set to 'formdata'

PR #863, opened 2024-08-02
Revert "Fix contentFormatter for FormData returning an empty FormData object"

smorimoto:
Reverts acacode/swagger-typescript-api#747


smorimoto:
Generally, the response is not in FormData format, and to use it we actually had to parse and reconstruct it as a FormData object, but I don't think that's necessary here.

PR #954, opened 2024-10-15
Respect silent option by only logging errors

### LINKED ISSUE #942: The option --silent is not honoured in version 13.0.22
It looks like the option --silent has no effect in version 13.0.22.

I verified the behaviour with the example yaml api-with-example.yaml (https://github.com/acacode/swagger-typescript-api/blob/main/tests/fixtures/schemas/v3.0/api-with-examples.yaml):

```bash
swagger-typescript-api --modular --module-name-first-tag --unwrap-response-data --axios -p ~/Downloads/api-with-examples.yaml -o /tmp --silent
```

The output in version 13.0.22 is

```
ℹ Code generator will use the default template for "api"                                                                                                                                                        
ℹ Code generator will use the default template for "routetypes"                                                                                                                                                 
ℹ try to get swagger by path "/home/marnij/Downloads/api-with-examples.yaml"                                                                                                                                    
ℹ start generating your typescript api                                                                                                                                                                          
✔ api file "http-client.ts" created in /tmp                                                                                                                                                                     
✔ api file "V2.ts" created in /tmp  
```

but in 13.0.16 there is no output (which is the closest previous version that I could get to work)

Regards,

Martin

Kiina:
Issue introduced by #903

Fixes #942

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -53,6 +53,9 @@ const generateTemplatesCommand = defineCommand({
     },
   },
   run: async ({ args }) => {
+    if (args.debug) consola.level = Number.MAX_SAFE_INTEGER;
+    if (args.silent) consola.level = 0;
+
     await generateTemplates({
       cleanOutput: args["clean-output"],
       httpClientType: args["http-client"],
@@ -270,6 +273,9 @@ const generateCommand = defineCommand({
     },
   },
   run: async ({ args }) => {
+    if (args.debug) consola.level = Number.MAX_SAFE_INTEGER;
+    if (args.silent) consola.level = 0;
+
     let customConfig = null;
     let customConfigPath: string | undefined;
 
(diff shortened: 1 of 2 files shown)


smorimoto:
It should be handled in the entry point: https://github.com/acacode/swagger-typescript-api/blob/main/src/index.ts

Kiina:
Adding it in the entrypoint didn't silence it when running:

`tsx index.ts generate-templates --silent -p ./tests/fixtures/schemas/v3.0/petstore.yaml -o ./src -n myApi.ts`

to test it, only when running:

`tsx src/index.ts generate --silent -p ./tests/fixtures/schemas/v3.0/petstore.yaml -o ./src -n myApi.ts`

So I added it there. But I can move it

smorimoto (approved):

Issue #965, opened 2024-10-23
When I lost many time on non working multi file upload, I was surprised why it's not supported. Moreover, fix is trivial, something like


ikzsl:
When I lost many time on non working multi file upload, I was surprised why it's not supported. Moreover, fix is trivial, something like

```ts
[ContentType.FormData]: (input: any) =>
  Object.keys(input || {}).reduce((formData, key) => {
    const property = input[key];
    if (Array.isArray(property) && property.every(p => p instanceof Blob)) {
      property.forEach(p => formData.append(key, p));
    } else {
      formData.append(
        key,
        property instanceof Blob
          ? property
          : typeof property === 'object' && property !== null
            ? JSON.stringify(property)
            : `${property}`,
      );
    }
    return formData;
  }, new FormData()),
```

_Originally posted by @darky in https://github.com/acacode/swagger-typescript-api/discussions/815#discussioncomment-10907960_

guatedude2:
I have a similar issue in React Native but I use `uri` with local system files so I added one more check:
```typescript
[ContentType.FormData]: (input: any) =>
  Object.keys(input || {}).reduce((formData, key) => {
    const property = input[key]
    if (Array.isArray(property) && property.every((p) => 'uri' in p)) {
      property.forEach((p) => formData.append(key, p))
    } else if (Array.isArray(property) && property.every((p) => p instanceof Blob)) {
      property.forEach((p) => formData.append(key, p))
    } else {
      formData.append(
        key,
        property instanceof Blob ? property : typeof property === 'object' && property !== null ? JSON.stringify(property) : `${property}`,
      )
    }
    return formData
  }, new FormData()),
```

Issue #705, opened 2024-04-26
Multipart formdata with file input gets wrapped in object when generated


martkaa:
This is a multipart form data with file input. I am trying to post images to a given trip id. I think there is an issue with how the multipart form data endpoint is generated:

What swagger-typescript-api generates:
``` 
tripImageCreate: (
            id: string,
            data: {
                tripImageFiles: File[]
            },
            params: RequestParams = {}
        ) =>
            this.request<TripModel, string>({
                path: `/api/trip/${id}/image`,
                method: 'POST',
                body: data,
                type: ContentType.FormData,
                format: 'json',
                ...params,
            }),
```

How it should be generated (this works):
```
 tripImageCreate: (
            id: string,
            data: File[],
            params: RequestParams = {}
        ) =>
            this.request<TripModel, string>({
                path: `/api/trip/${id}/image`,
                method: 'POST',
                body: data,
                type: ContentType.FormData,
                format: 'json',
                ...params,
            }),
```

This is the swagger.json for the endpoint that works fine when i post image files in the swagger gui. 
```
"/api/trip/{id}/image": {
      "post": {
        "tags": [
        XX
        ],
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "required": true,
            "style": "simple",
            "schema": {
              "type": "string",
              "format": "uuid"
            }
          }
        ],
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "required": [
                  "tripImageFiles"
                ],
                "type": "object",
                "properties": {
                  "tripImageFiles": {
                    "type": "array",
                    "items": {
                      "type": "string",
                      "format": "binary"
                    }
                  }
                }
              },
              "encoding": {
                "tripImageFiles": {
                  "style": "form"
                }
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "Success",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/TripModel"
                }
              }
            }
          },
          "400": {
            "description": "Bad Request",
            "content": {
              "application/json": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        }
      }
    },
```

PR #1008, opened 2024-12-11
Fix handling of input of type FormData for the Fetch client

frazar:
Several open issues are caused by the `multipart/form-data` formatter trying to provide a simple API based on an input of type `Object`. However, `Object` can not express some of the typical use cases of a multipart request. On the other hand, `FormData` could.

One example (mentioned in #705 and #965) is sending a requested with multiple entries belonging to the same "key". Since a `FormData` object allows one key to be mapped to multiple values, one can simply do:

```ts
f = new FormData()
f.append('files', new Blob())
f.append('files', new Blob())
[...f.keys()] // This gives 2 entries, as expected
```

This is clearly not possible with an `Object`, where the key-value relationship is 1:1. #1000 tries to achieve this by automatically converting `Array` values to multiple `FormData` keys. However, I argue that it is not necessarily what the user wants, since they could either expect:
- that the array is sent as-is as a single `FormData` entry, or
- that each array is sent a separate `FormData` entries.

In principle, it is not so easy to infer which of the two is the one desired by the user.

Currently, however, inputs of type `FormData` are not correctly handled by the `multipart/form-data` formatter for the Fetch client, which tries to enumerate the keys using the `Object.keys(...)` construct:

https://github.com/acacode/swagger-typescript-api/blob/325e308fee3d8e6c0e2fc3e6d3bb2328c8725716/templates/base/http-clients/fetch-http-client.ejs#L107-L108

Unfortunately, `FormData` returns an empty list when used through `Object.keys()`

```ts
f = new FormData()
f.append('files', new Blob())
f.append('files', new Blob())
Object.keys(f) // This gives an empty array!
```

This PR proposes a way to allow the user to directly supply a `FormData`, basically opting out of any smart formatting logic that might actually get in the way. In this way, the user takes responsibility for correctly managing the conversion.

I argue that the breaking change is minimal since any use of `FormData` is broken under the current API. Additionally, this would align with the logic for the Axios client, which already short-circuits the formatting for `FormData` inputs.

https://github.com/acacode/swagger-typescript-api/blob/325e308fee3d8e6c0e2fc3e6d3bb2328c8725716/templates/base/http-clients/axios-http-client.ejs#L81-L82

### DIFF
--- a/.changeset/loud-dancers-shake.md
+++ b/.changeset/loud-dancers-shake.md
@@ -1,0 +1,19 @@
+---
+"swagger-typescript-api": patch
+---
+
+Fix handling of FormData inputs in Fetch HTTP client
+
+Previously, when users passed a `FormData` object directly to the Fetch
+client's `multipart/form-data` formatter, it would incorrectly attempt to use
+`Object.keys()` on the FormData instance, which returns an empty array. This
+caused the FormData to be processed incorrectly.
+
+The fix adds a type check to return FormData instances unchanged, allowing
+users to have full control over FormData construction when needed whilst
+maintaining backwards compatibility for object inputs. This aligns the Fetch
+client behaviour with the existing Axios client implementation.
+
+This resolves issues where users needed to send multipart requests with
+multiple entries for the same key, which is only possible with direct FormData
+manipulation.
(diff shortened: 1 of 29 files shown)


DmitryShcheklein:
How long will this fix stay open? The problem is actual 😤

PavelPopovIPS:
Hi everyone, resolving this issue can be helpfull for nodejs projects.
Is there chance to check and approve changes?

antonkharkhonov:
+1, I encountered the same problem. So far, I have been adding this early return manually each time after generating a new build. Therefore, it would be very helpful if this could be checked and merged.

smorimoto (approved):

PR #1041, opened 2025-01-31
Add ts-nocheck comment to auto-generated file

Jerome1337:
This is useful when the openapi contains unused variables.
Otherwise the TSC type checking crash on auto generated code...


Jerome1337:
I don't understand why tests aren't passing. This seems do be indentation differences between snapshots and expectation.

smorimoto (approved):

PR #1064, opened 2025-02-22
Fix route param parsing when the param begins with just one letter (rebased original PR #686 by @ahoseinian)

nicky1038:
Fixes the way how query params are parsed when they consist of one letter or the first letter is followed by underscore.

Original #686 by @ahoseinian was needed to be rebased.

So I did it myself because of no activity there.

### DIFF
--- a/.changeset/honest-bears-stay.md
+++ b/.changeset/honest-bears-stay.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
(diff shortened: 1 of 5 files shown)


smorimoto (approved):

smorimoto:
@nicky1038 Let me know what you think about the last commit. The last commit feels fine to me.

smorimoto (approved):

PR #1114, opened 2025-03-24
Added processing of array of properties

FanToMaS-api:
Added processing of array of properties, for adding to formData sequentially, element by element

I Have the problem 

my backend contract is

```csharp
/// <summary>
///     Новые фотографии автосервиса
/// </summary>
[MaxFileSize(ImageValidationConstants.MaxFileSize)]
[AllowedExtensions([".png", ".jpg", ".jpeg", ".heic"])]
public IFormFile[]? NewPhotos { get; set; }

/// <summary>
///     Фотографии, которые надо удалить
/// </summary>
public int[]? PhotoDeletionIds { get; set; }
```

when i send request from front 

```typescript

export interface UpdateCarServicePhotosPayload {

  /** Новые фотографии автосервиса */
  newPhotos?: File[];
  
  /** Фотографии, которые надо удалить */
  photoDeletionIds?: number[];
}
```

newPhotos is array of File. But in client.ts by swaggergen I get an attempt to add the entire array under one key as a single blob object

the code below doesn't work good with array of objects

```typescript
               formData.append(
                  key,
                  property instanceof Blob
                    ? property
                    : typeof property === "object" && property !== null
                      ? JSON.stringify(property)
                      : `${property}`
                );
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "swagger-typescript-api",
-  "version": "13.0.26",
+  "version": "13.0.27",
   "description": "Generate the API client for Fetch or Axios from an OpenAPI Specification",
   "homepage": "https://github.com/acacode/swagger-typescript-api",
   "bugs": "https://github.com/acacode/swagger-typescript-api/issues",


smorimoto:
bugbot run

PR #1167, opened 2025-04-10
Support v3.1 typed arrays

### LINKED ISSUE #991: Nullable array with `$ref` is generated as `any[] | null`
I have an object with the following (non-required) property defined in the schema (OpenAPI 3.1.0):

```json
"foos": {
  "type": ["array", "null"],
  "items": {
    "$ref": "#/components/schemas/Foo"
  }
}
```

On 13.0.22, this generates: `foos?: any[] | null`

I would instead expect: `foos?: Foo[] | null`

The same thing happens if the property is required, as well. If I make it non-nullable, it works as expected.

Thomasdezeeuw:
In OpenAPI v3.1 the nullable property was replaced with a "typed array", that is an array that can either be of type T or null.

To make the minimal amount of changes this commit converts a type array back into what a 3.0 parser would expect.

Closes #991

### DIFF
--- a/src/schema-parser/base-schema-parsers/object.ts
+++ b/src/schema-parser/base-schema-parsers/object.ts
@@ -36,6 +36,17 @@ export class ObjectSchemaParser extends MonoSchemaParser {
         "rawTypeData",
         {},
       );
+      // In OpenAPI v3.1 the nullable property was replaced with a "typed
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
@smorimoto any chance you could review this? We could really use this change.

PR #1260, opened 2025-05-30
fix: keep underscores in enum names with numbers (like ENUM_123_VALUE)

### LINKED ISSUE #276: Broken/unexpected behavior for enum values with numbers in them
When generating enums (without `--union-enums`), there are two things that can happen when there are numbers in the enum values:

1. An enum value prefixed with underscore and a number will result in invalid code being generated. For example `_123_ENUM_VALUE` in the swagger enum will generate the name `123ENUMVALUE` which is invalid typescript because it starts with a number. Expected name would probably be to keep `_123_ENUM_VALUE` on the typescript side as well.

2. An enum value with numbers anywhere in it is has underscores stripped from the generated name (reminiscent of #108). For example `ENUM_123_VALUE` in the swagger enum will generate the name `ENUM123VALUE`. Expected typescript name would be `ENUM_123_VALUE` with underscores intact because it's valid typescript as-is.

parsaghk:
Hi :wave: 

Before, enum names like ENUM_123_VALUE lost their underscores and became ENUM123VALUE.  
This change keeps the underscores so the names stay correct. (Fixes #276)

Best,
Parsa

### DIFF
--- a/tests/spec/enumIncludesNumber/schema.json
+++ b/tests/spec/enumIncludesNumber/schema.json
@@ -1,0 +1,13 @@
+{
+  "openapi": "3.0.0",
+  "tags": [],
+  "servers": [],
+  "components": {
+    "schemas": {
+      "StringEnumWithoutKey": {
+        "type": "string",
+        "enum": ["VAL_1", "VAL_2", "VAL_3", "_1_VAL"]
+      }
+    }
+  }
+}
--- a/tests/spec/enumIncludesNumber/schema.json
+++ b/tests/spec/enumIncludesNumber/schema.json
@@ -4,9 +4,17 @@
   "servers": [],
   "components": {
     "schemas": {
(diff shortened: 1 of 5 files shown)


tests/spec/enumIncludesNumber/schema.json
```diff
@@ -0,0 +1,13 @@
+{
+  "openapi": "3.0.0",
+  "tags": [],
+  "servers": [],
+  "components": {
+    "schemas": {
+      "StringEnumWithoutKey": {
+        "type": "string",
+        "enum": ["VAL_1", "VAL_2", "VAL_3", "_1_VAL"]
```
smorimoto, line 9: Could you add more test cases like `A_1_B_2_C`?

smorimoto:
Could you please add changeset as well?

Issue #1040, opened 2025-01-30
Generate objects instead of enums


scottcorgan:
With Typescript 5.8's `erasableSyntaxOnly` flag and with Node's type stripping support, it looks like they're pushing us away from from using TS only language features (enum, namespaces, etc) We heavily rely on the enums as generated objects for typesafety everywhere.

Would love to be able to just generate the object version of these enums in the generated output.

nathan-de-pachtere:
Tried with that template but can't make it works, the generate process don't use it at all ... others templates are used correctly but this one not. 

```ejs
<%
const { contract, utils, config } = it;
const { formatDescription, require, _ } = utils;
const { name, $content } = contract;
%>
<%_
  const enumName = name.toUpperCase();
  const enumValues = Object.entries($content);
_%>
<% if (config.generateUnionEnums) { %>
  export type <%~ name %> = <%~ _.map($content, ({ value }) => value).join(" | ") %>
<% } else { %>
  export const <%= enumName %>: { [key: string]: string } = {
    <%_ enumValues.forEach(([key, value]) => { _%>
      <%= key %>: '<%= value %>',
    <%_ }); _%>
  } as const;

// export type name = ObjectValues<typeof enumName>; // didn't make it to the step of generating actual type from that const but that the goal.
<% } %>
```

Also tried with 
```typescript
codeGenConstructs: (struct) => {
        return {
          ...struct,
          EnumField: ... //Can't figure out how to modify that as the doc seems not up to date
        };
      },
```

@scottcorgan did you find a way ?

nathan-de-pachtere:
@scottcorgan https://github.com/OpenAPITools/openapi-generator 
This tool make it by default ;)

quovide:
Welcoming feature indeed

Upgrade220:
Introduced another PR cause found previous one made by @bjohnso5 not complete. All reasons explained inside PR.

PR #1278, opened 2025-06-09
feat: Support generating erasable syntax enum-style types

bjohnso5:
First stab at supporting #1040 via a new CLI flag. It currently takes lower precedence than generateUnionEnums - so if both are set, unions will always be written out.

I'm not 100% confident I've fully grokked the intended structure of the added code, so please point out any deficiencies that have been added, and I'll try to correct them.

Cheers!


bjohnso5:
This needs more time to cook - I'm going to close this for now and work on it before re-opening.

PR #1281, opened 2025-06-10
feat: const object-style enum generation

### LINKED ISSUE #1040: Generate objects instead of enums

bjohnso5:
Closes #1040.

### DIFF
--- a/.changeset/honest-feet-worry.md
+++ b/.changeset/honest-feet-worry.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": minor
+---
+
+Implement const object-style enum generation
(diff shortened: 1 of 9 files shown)


bjohnso5:
I'm not clear on how best to extend this to the inline `enum` generation, or if that even makes sense - I'd love some guidance.

wes-goulet (approved):

smorimoto:
bugbot run

scr2em:
Any chance this gets merged?

Issue #532, opened 2023-05-22
How do I handle a colon in the url? 


cfluke-cb:
I have an http path like this
"/v1/{walletName}:generateAddress": 

The :generateAddress is becoming a string parameter from the generated client.

`
const response = await client.walletServiceGenerateAddress(name, ':generateAddress', {
    network: networkName,
 });
`

Is there a way to escape the colon in the path? or an generate parameter to ignore colon based http path variables?

Moloko20:
Do you find the solution?

I get same error :(

PR #1284, opened 2025-06-12
fix regex related parsing route name

ghg070:
We used swagger-typescript-api to parse and encountered the same issue as this developer (https://github.com/acacode/swagger-typescript-api/issues/532). Is it possible to modify the regex in parseRouteName to ensure that there is always a "/" or "..." at the beginning before treating it as a parameter?


smorimoto:
bugbot run

PR #1350, opened 2025-08-05
feat: add immutable types

J3MPA:
In a functional codebase consumers of `swagger-typescript-api` would want to force properties and values to be immutable even if swagger spec does not have `readOnly` property.

With this change one can simply add `--immutable` flag or set `immutable` option to force immutable types.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -58,7 +58,7 @@ export class CodeGenConfig {
   /** CLI flag */
   addReadonly = false;
   /** CLI flag */
-  makeImmutable = false;
+  immutable = false;
   enumNamesAsValues = false;
   /** parsed swagger schema from getSwaggerObject() */
 
--- a/.changeset/real-wombats-dream.md
+++ b/.changeset/real-wombats-dream.md
@@ -2,4 +2,4 @@
 "swagger-typescript-api": minor
 ---
 
-Supports immutable types with --make-immutable or makeImmutable
+Supports immutable types with --immutable flag or immutable option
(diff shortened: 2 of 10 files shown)


J3MPA:
@smorimoto I'm a little bit unsure on when to use lodash or not in the lib, seems like there is no clear pattern on when it's applied or not. 

Do you have time to review the PR?

smorimoto:
bugbot run

J3MPA:
@smorimoto I fixed the suggestion from cursor

J3MPA:
@smorimoto Could these changes be approved, merged and released?

smorimoto:
bugbot run

J3MPA:
@smorimoto Any updates, can we aim to have this merged?

smorimoto:
I will take a look at this today. Sorry for being late. Also, how about simply making the flag “immutable”? I think it's simpler.

J3MPA:
> I will take a look at this today. Sorry for being late. Also, how about simply making the flag “immutable”? I think it's simpler.

Yes I agree, i tried following naming convention of other flags, hence the "make" part. But "immutable" is for sure better, will update it

J3MPA:
@smorimoto Any updates? :)

src/configuration.ts
```diff
@@ -344,8 +370,11 @@ export class CodeGenConfig {
     /**
      * [$A1, $A2, ...$AN]
      */
-    Tuple: (values: unknown[]) => {
-      return `[${values.join(", ")}]`;
+    Tuple: ({
+      readonly,
+      values,
+    }: Record<string, unknown> & { values: unknown[] }) => {
```
js2me: Better to safe backward compatibility with this function. Just add second arg as extra options with optional property readOnly?: boolean
I mean

values:unknown[], opts?:{ readOnly?: boolean}

PR #1364, opened 2025-08-16
fix: ensure ext parameter includes leading dot in path formatting

ava-grace-zoe:
- Explicitly prepend '.' to ext if missing (e.g., "ts" → ".ts")


smorimoto:
bugbot run

PR #1375, opened 2025-08-26
feat: Add opt-in JSON-LD schema support

dev-geddy:
Adds optional JSON-LD support. Off by default, no effect on existing specs.

- Enable with `--jsonld` (or `jsonLdOptions.enabled`)
- Schemas opt in with the `x-jsonld` extension
- Entities get typed `@context`, `@type`, `@id` and extend a shared `JsonLdEntity`
- A `x-jsonld-type` schema with no properties becomes a string union alias
- Modular output adds `jsonld-entity` and `jsonld-utils`, re-exported from `data-contracts`
- `generateUtils: false` emits plain interfaces with no shared module

## Problem

If your API returns JSON-LD, the generated types don't mean much. `@context` becomes `object`, nothing connects the JSON-LD fields across entities, and there are no shared types for things like `@graph` or node references:

```ts
export interface Person {
  "@context"?: object;
  "@type": "Person" | "https://schema.org/Person";
  "@id"?: string;
  name: string;
}
```

So you end up writing those types by hand next to the generated file.

## Solution

Two things must both be true for anything to change: the schema declares `x-jsonld` (or `x-jsonld-context` / `x-jsonld-type` / `x-jsonld-id`), and you pass `--jsonld`. Otherwise output is identical to before.

Under the hood this adds two schema types, `jsonld-entity` and `jsonld-type`, as `MonoSchemaParser` subclasses dispatched from `getInternalSchemaType` — the same way `enum`, `object` and `array` already work. No new pipeline stage.

An entity now looks like:

```ts
import type { JsonLdContext, JsonLdEntity } from "./jsonld-utils";

/** JSON-LD Entity: Person */
export interface Person extends JsonLdEntity {
  "@context"?: JsonLdContext;
  "@type": "https://schema.org/Person";
  "@id"?: string;
  name: string;
}
```

And a type-only schema becomes an alias, staying in `data-contracts` with the other aliases:

```ts
export type PersonType = "Person" | "Patient";
```

A few choices worth flagging:

- **Detection is explicit.** I didn't auto-detect from `@context`/`@type`/`@id` property names, since those are valid property names in a normal spec and it would change output for people who never asked for it.
- **The input document is never modified.**
- **Only entities move to their own file** in modular mode, and `data-contracts` re-exports them. Route modules import models from `data-contracts` only, so without the re-export their imports would break.
- **`@context` uses `JsonLdContext`** rather than a structural type. `object` or `(string | object)[]` isn't assignable to `JsonLdEntity["@context"]`, so the generated file wouldn't compile. Literal string contexts keep their literal type.

## Verification

`bun run build`, `bun run format:check` and `bun run test` pass — 293 tests, no type errors.

`tests/spec/jsonld-basic/` covers single-file output, the disabled default, modular output (including that route module imports still resolve), the `generateUtils: false` path, a mixed JSON-LD/plain spec, and the type alias case.

I also compiled the real generated output rather than only snapshotting it: for all three modes the emitted files pass `tsc --noEmit --strict` with the generated `// @ts-nocheck` header removed, so the check isn't suppressed.

## Notes

Rebased onto current `main` and squashed to one commit. The rebase surfaced that the project had moved to `es-toolkit`, which this now uses.

An earlier version had a `generateContext` option with its own parser and template. They're gone: nothing could produce a `jsonld-context` schema type, so the option did nothing and the code was unreachable. Happy to build a real context output if that's wanted, but it should be deliberate rather than dead code.


smorimoto:
bugbot run

smorimoto:
bugbot run

smorimoto:
@codex review

PR #1381, opened 2025-08-28
Handle extraction of path-only request parameters

### LINKED ISSUE #1323: --extract-request-params on DELETE method
It should extract the params on DELETE methods also if I provide `--extract-request-params` in cli.
Currently works on other methods like GET, POST, PATCH ...

Image

smorimoto:
## Summary
- support extracting request parameters when an operation only defines path parameters
- update snapshots accordingly

## Testing
- `yarn test`

------
https://chatgpt.com/codex/tasks/task_e_68b008db0c308327b6a1e19eff065502

### DIFF
--- a/.changeset/extract-request-params-delete.md
+++ b/.changeset/extract-request-params-delete.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
+Support extracting request parameters for routes that only declare path parameters, such as DELETE methods.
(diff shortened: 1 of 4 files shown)


Khavrolev:
Agree with @porfirioribeiro. Also that's a breaking change, you don't expect it from patch bump

yelvert:
This is a major breaking change coming from AI slop. It should never have been merged as a patch release if at all.

PR #1393, opened 2025-09-08
Set discriminators types before other types than enums

RoCat:
The goal of this merge request is to ensure discriminators are at the (not so) very first place in generation, just after enums.
If they're not, that will cause cyclic errors in typescript.

### DIFF
--- a/src/code-gen-process.ts
+++ b/src/code-gen-process.ts
@@ -127,6 +127,9 @@ export class CodeGenProcess {
       }),
     );
 
+    //set all discriminators at the top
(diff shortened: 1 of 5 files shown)


src/code-gen-process.ts
```diff
@@ -127,6 +127,9 @@ export class CodeGenProcess {
       }),
     );
 
+    //set all discriminators at the top
```
smorimoto, line 130:
```suggestion
    // set all discriminators at the top
```

Issue #1366, opened 2025-08-18
Version 13.2.8 generates incorrect TS


buffet-time:
# Issue in version 13.2.8

## Context

I'm working on an [open source project](https://github.com/leaderboardsgg/leaderboard-site) and we have a [C# backend](https://github.com/leaderboardsgg/leaderboard-backend) using `Swagger` and so of course we use this wonderfully helpful package for our development. I am in the process of updating all of our packages and I ran into the issue where bumping this to `13.2.8` causes errors preventing builds.

## Errors:

```txt
// This is an error appearing on the `.` of `this.request` see below
Expected '=' for property initializer.ts(1442) 

// on the `|` of the same line
',' expected.ts(1005)

// on the closing `>` 
Expression expected.ts(1109)
```

## generated code
Below is the unmodified output from the generated `Leaderboards.ts` for `13.2.7` and `13.2.8` 

### `13.2.7` Working correctly:
```ts
/* eslint-disable */
/* tslint:disable */
// @ts-nocheck
/*
 * ---------------------------------------------------------------
 * ## THIS FILE WAS GENERATED VIA SWAGGER-TYPESCRIPT-API        ##
 * ##                                                           ##
 * ## AUTHOR: acacode                                           ##
 * ## SOURCE: https://github.com/acacode/swagger-typescript-api ##
 * ---------------------------------------------------------------
 */

import {
  CreateLeaderboardRequest,
  GetLeaderboardBySlugParams,
  LeaderboardViewModel,
  ListLeaderboardsParams,
  ProblemDetails,
  UpdateLeaderboardRequest,
  ValidationProblemDetails,
} from "./data-contracts";
import { ContentType, HttpClient, RequestParams } from "./http-client";

export class Leaderboards<
  SecurityDataType = unknown,
> extends HttpClient<SecurityDataType> {
  /**
   * No description
   *
   * @tags Leaderboards
   * @name GetLeaderboard
   * @summary Gets a leaderboard by its ID.
   * @request GET:/api/leaderboard/{id}
   * @secure
   * @response `200` `LeaderboardViewModel` OK
   * @response `400` `ProblemDetails` Bad Request
   * @response `404` `void` Not Found
   * @response `500` `void` Internal Server Error
   */
  getLeaderboard = (id: number, params: RequestParams = {}) =>
    this.request<LeaderboardViewModel, ProblemDetails | void>({
      path: `/api/leaderboard/${id}`,
      method: "GET",
      secure: true,
[...]

PR #1398, opened 2025-09-10
Add test that generates content for modular

k1rd3rf:
This test will reproduce the issues in #1366.

The snapshot is based on based on version 13.2.7.


k1rd3rf:
image

PR #1430, opened 2025-09-23
fix: ESM compatibility - avoid direct Response object mutation

### LINKED ISSUE #1429: Fix: Response object mutation fails in ESM environments with node-fetch v3
## Problem
The generated Fetch client attempts to directly mutate the Response object by adding `data` and `error` properties. This fails in ESM environments using node-fetch v3 because Response objects are read-only.

## Error
```
TypeError: Cannot set property data of #<Body> which has only a getter
    at /node_modules/.pnpm/@oars+roleauth-api@1.0.53765/node_modules/@oars/v1.ts:1237:13
```

## Environment
- Node.js with ESM modules (`"type": "module"` in package.json)
- node-fetch v3.x (ESM-only version)
- swagger-typescript-api v13.0.23 (also confirmed in v13.2.13)

## Root Cause
In `templates/base/http-clients/fetch-http-client.ejs` (lines 187-189), the code attempts to directly mutate the Response object:

```javascript
const r = response as HttpResponse<T, E>;
r.data = (null as unknown) as T;
r.error = (null as unknown) as E;
```

This works in CommonJS environments but fails in ESM with node-fetch v3 where Response objects are truly read-only and cannot have new properties added.

## Impact
Anyone migrating to ESM modules or using node-fetch v3 cannot use the generated Fetch client without workarounds.

## Current Workaround
Users must wrap fetch with a Proxy to make Response properties writable:

```javascript
const wrappedFetch = async (...args) => {
  const response = await fetch(...args);
  return new Proxy(response, {
    set(target, prop, value) {
      if (prop === 'data' || prop === 'error') {
        target[prop] = value;
        return true;
      }
      return Reflect.set(target, prop, value);
    },
    get(target, prop) {
      if (prop === 'data' || prop === 'error') {
        return target[prop];
      }
      const value = target[prop];
      if (typeof value === 'function') {
        return value.bind(target);
      }
      return value;
    }
  });
};
```

## Proposed Solution
Create a wrapper object instead of mutating the Response directly. This maintains backward compatibility while supporting modern ESM environments.

I'll be submitting a PR with a fix shortly.

omarmciver:
## Summary
Fixes #1429 - Response object mutation fails in ESM environments

## Problem
The Fetch client template directly mutates Response objects by adding `data` and `error` properties. This works in CommonJS but fails in ESM with node-fetch v3 where Response objects are truly read-only.

## Solution
Instead of mutating the Response object, this PR creates a wrapper object that:
- Contains the custom `data` and `error` properties
- Delegates all Response properties and methods to the original response
- Maintains full backward compatibility with existing code
- Works in both CommonJS and ESM environments

## Changes
Modified `templates/base/http-clients/fetch-http-client.ejs` to create a wrapper object instead of directly mutating the Response:

```javascript
// Before (fails in ESM):
const r = response as HttpResponse<T, E>;
r.data = null;
r.error = null;

// After (works everywhere):
const r = {
  data: null,
  error: null,
  // Delegate Response properties
  ok: response.ok,
  status: response.status,
  // ... etc
  // Delegate Response methods
  json: () => response.json(),
  text: () => response.text(),
  // ... etc
} as HttpResponse<T, E>;
```

## Testing
- ✅ Created test script to verify the template generates correct code
- ✅ Tested that the fix doesn't directly mutate Response objects
- ✅ Verified wrapper object is created properly
- ✅ Confirmed backward compatibility is maintained

## Breaking Changes
None - this change is fully backward compatible. The generated API clients will have the same interface and behavior, but will now work in ESM environments.

## Additional Notes
This fix enables users to migrate to:
- ESM modules (`"type": "module"` in package.json)
- node-fetch v3 (ESM-only version)
- Native fetch in Node.js 18+

Without this fix, users need workarounds like wrapping fetch with a Proxy to make Response properties writable, which is not ideal.

### DIFF
--- a/templates/base/http-clients/fetch-http-client.ejs
+++ b/templates/base/http-clients/fetch-http-client.ejs
@@ -203,7 +203,7 @@ export class HttpClient<SecurityDataType = unknown> {
                 get(target, prop) {
                     // Custom properties for our API wrapper
                     if (prop === 'data') {
-                        return target._data !== undefined ? target._data : (null as unknown) as T;
+                        return target._data ?? (null as unknown as T);
                     }
                     if (prop === 'error') {
                         return target._error !== undefined ? target._error : (null as unknown) as E;


smorimoto:
bugbot run

smorimoto:
@codex review

smorimoto:
@omarmciver Could you add a changeset?

PR #1431, opened 2025-09-23
fix: handle arrays properly in FormData

omarmciver:
## Description

This PR fixes an issue with FormData handling when arrays are involved, particularly for multi-file uploads.

## The Problem

When an API endpoint accepts an array of files (or any array in FormData), the current implementation incorrectly converts the entire array to a string representation instead of properly appending each item individually to the FormData object.

### Example scenario:
```javascript
// API expects multiple file attachments
const files = [file1, file2, file3];
api.uploadMultiple({ files });

// Current behavior: FormData contains
// "files" => "[object File],[object File],[object File]" ❌

// Expected behavior: FormData should contain
// "files" => file1
// "files" => file2  
// "files" => file3 ✅
```

## The Solution

This fix modifies the FormData content formatter to:
1. Detect when a property value is an array
2. Iterate through each item in the array
3. Properly append each item individually to FormData
4. Preserve the correct handling of Blob/File types vs other data types

## Changes Made

- Updated `templates/base/http-clients/fetch-http-client.ejs` to handle arrays in FormData
- Added test fixture `formdata-arrays.json` to verify the fix
- All existing tests pass with updated snapshots

## Use Cases

This fix is essential for:
- Multi-file upload endpoints
- Batch document processing APIs  
- Any FormData field that accepts multiple values
- Services handling attachments (e.g., agreement management systems)

## Testing

- ✅ All existing tests pass
- ✅ Snapshots updated to reflect the new array handling
- ✅ Added test fixture for multi-file upload scenario

## Backwards Compatibility

This change is fully backwards compatible:
- Single values continue to work as before
- Arrays are now handled correctly instead of being stringified
- No breaking changes to the API


smorimoto:
@codex review

Issue #314, opened 2021-11-05
Not generating unions with multiple media types


slinkardbrandon:
There seems to be an issue generating union types when multiple media types exist that alter the requirements for the request body. 

Example path with multiple media types ([right from the swagger requestBody docs](https://swagger.io/docs/specification/describing-request-body/)):

Given the following schema:
```yaml
paths:
  /pets:
    post:
      summary: Add a new pet
      requestBody:
        content:
          "application/json":
            schema:
              $ref: '#/components/schemas/Cat'
          "application/foo.bar+json":
            schema:
              $ref: '#/components/schemas/Dog'
```

Only one output method will be generated, it seems to either be `application/json` or index 0, I'm not entirely sure.

Is there any way I can get this to generate a union type rather than always generating from schemas/Cat?

markusgeert:
Also experiencing this issue, but when different response types are specified.

PR #1462, opened 2025-10-22
fix: union when there are multiple request/response types defined

stefan-cooper:
Contributes to #314

Signed-off-by: Stefan Cooper <stefan.cooper27@gmail.com>


skyless:
@stefan-cooper kind reminder :)

PR #1550, opened 2025-12-10
fix: nullable not respected for objects with nullable properties #533

### LINKED ISSUE #533: `nullable: true` ignored if `type: object` and a property has `nullable: true`
# What's wrong?
Hello again. I have a situation where I had an object with `nullable: true` and all was working well, then I needed to add some properties that had `nullable: true` and for some reason the ` | null` on the main object was removed from the types.

__This Schema:__
```yaml
MyDomainObjectSchema:
  required:
    - user
  properties:
    user:
      type: object
      nullable: true
      required:
        - id
        - displayName
        - email
      properties:
        id:
          type: string
        displayName:
          type: string
        email:
          type: string
          nullable: true
```

__Produces the following incorrect type:__
```typescript
export interface MyDomainObjectSchema {
  user: {
    id: string;
    displayName: string;
    email: string | null;
  }; // where is the union with `null`?
}
```

## What works

If there are no properties that are nullable then the parent object is correctly typed as shown below:
__This Schema:__
```yaml
MyDomainObjectSchema:
  required:
    - user
  properties:
    user:
      type: object
      nullable: true
      required:
        - id
        - displayName
        - email
      properties:
        id:
          type: string
        displayName:
          type: string
        email:
          type: string
          # removing `nullable: true` here fixes the incorrect behavior
          # nullable: true 
```
__Produces the following correct result:__

```typescript
export interface MyDomainObjectSchema {
  user: {
    id: string;
    displayName: string;
    email: string;
  } | null;
}
```

# What is expected

I expect that `nullable: true` will always apply to an object schema regardless of the properties and their conditions. I expect the original YAML schema provided would produce the following type:

```typescript
export interface MyDomainObjectSchema {
  user: {
    id: string;
    displayName: string;
    email: string | null;
  } | null;
}
```

Thanks for the hard work on this. We're still happy to have it to work with :)

bankyadam:
# Summary

Fixes #533 - Resolves incorrect null handling for nullable parent objects containing nullable child properties.

# Problem

When a schema defined a nullable object (`type: object, nullable: true`) with child properties that were also nullable, the generator was incorrectly adding an additional `| null` to the parent type. This happened because `isNullMissingInType` was detecting the `| null` from nested properties and incorrectly concluding that the parent type needed null added.

## Example Issue

Given this schema:
```json
{
  "type": "object",
  "nullable": true,
  "properties": {
    "email": { "type": "string", "nullable": true }
  }
}
```
Before: Generated redundant `| null` even though the parent was already properly nullable
After: Correctly generates `{ email?: string | null } | null`

# Solution

Refactored the `isNullMissingInType` method in `src/schema-parser/schema-utils.ts` to distinguish between:
  - Root-level nullable types (union with `null` at the top level)
  - Types with nested nullable properties (objects containing prop: `string | null`)

The fix uses regex patterns to detect only root-level null unions:
  - Pattern ending with `| null`
  - Pattern starting with `null |`
  - Exact match of `null`

This prevents false positives from nested nullable properties while correctly identifying when the parent type actually needs `| null` added.

# Changes

##  Core Fix

  - `src/schema-parser/schema-utils.ts`: Improved `isNullMissingInType` logic with regex-based root-level null detection

## Build Configuration

  - `package.json`: Changed ESM output extensions from .js to .mjs for better module resolution

##  Test Coverage

  - Added new test suite: `tests/spec/nullable-parent-with-nullable-children/`
  - Tests cover nullable parent objects with nullable child properties, nested nullable objects, and complex scenarios
  - Updated snapshots across multiple test suites to reflect correct nullable handling

#  Testing

  All existing tests pass with updated snapshots showing correct nullable type generation. New test suite specifically validates the fix for nested nullable scenarios.

---

> [!NOTE]
> Fixes null handling by only adding `| null` at the root type level, adds targeted tests, and updates snapshots accordingly.
> 
> - **Core Fix**
>   - Refactor `src/schema-parser/schema-utils.ts:isNullMissingInType` to detect only root-level `null` in unions (regex-based), avoiding false positives from nested nullable properties.
>   - `safeAddNullToType` now leverages the corrected check to prevent duplicate `| null` on parent object types.
> - **Tests**
>   - Add `tests/spec/nullable-parent-with-nullable-children/` suite with schema and snapshot validating nullable parent + nullable children behavior.
>   - Update snapshots (`tests/__snapshots__/extended.test.ts.snap`, `tests/__snapshots__/simple.test.ts.snap`, `tests/spec/additional-properties-2.0/__snapshots__/basic.test.ts.snap`) to reflect corrected nullability (e.g., unions with `null`, nullable record values).
> - **Meta**
>   - Add changeset `.changeset/giant-hats-work.md` (patch).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6db96156cd5b5e1c6cac7cba35f7762efb777c63. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


blinikar:
Hello, @bankyadam. I am writing to check whether you are able to resolve merge conflicts in your PR? It seems like you need to reabse only one small code snippet and it won't take long time.

If not, maybe I can rebase this PR for you?

smorimoto:
@cubic-dev-ai

PR #1603, opened 2026-02-07
Fix return type syntax in describeReturnType function

clicktodev:
Correct the return type syntax in the `describeReturnType` function to ensure proper TypeScript type definitions.


smorimoto:
@codex review

PR #1626, opened 2026-02-21
fix: preserve OpenAPI spec property order when sortTypes is false (#151)

### LINKED ISSUE #151: Any way to keep interface property in the same order
Hey,

We commit the generated API code to follow change through our stories. It's good but the order of each interface change at each generation.

First generation:
```
interface A {
  b: string;
  c: string;
}
```
second generation:
```
interface A {
  b: string;
  c: string;
}
```

eduardbar:
## Summary

Fixes #151.

When `--sort-types` is **not** set (the default), generated TypeScript interface properties did not maintain the same order as defined in the OpenAPI/Swagger specification. This made it hard to compare the generated output against the spec and caused unnecessary diffs in code reviews when the spec was updated.

## Root Cause

`SchemaComponentsMap.enumsFirst()` and `discriminatorsFirst()` used `Array.prototype.sort()` to promote enum and discriminator components to the front of the list. While `.sort()` is stable in V8 (and therefore preserves relative order for equal elements), the comparator returned `0` for non-matching items, which in theory allows the engine to move them. More critically, the *two* sorts are run back-to-back:

```ts
this.schemaComponentsMap.discriminatorsFirst(); // sort 1
this.schemaComponentsMap.enumsFirst();           // sort 2
```

After the second sort, non-enum components that are not discriminators may end up in an order determined by the sort algorithm's internal state rather than the spec's declaration order. In practice this was observable in projects with a mix of enums, discriminators, and plain object schemas.

## Fix

Replace the sort-based implementation with a **partition-based stable reorder** via a new private `_stablePromoteToFront()` helper:

```ts
private _stablePromoteToFront(predicate: (c: SchemaComponent) => boolean) {
  const promoted: SchemaComponent[] = [];
  const rest: SchemaComponent[] = [];
  for (const component of this._data) {
    if (predicate(component)) promoted.push(component);
    else rest.push(component);
  }
  this._data = [...promoted, ...rest];
}
```

This is O(n) and **guaranteed stable** — the relative order of every component that does not match the predicate is preserved exactly. Both `enumsFirst()` and `discriminatorsFirst()` now delegate to this helper.

The `--sort-types` path (alphabetical sort in `code-gen-process.ts` and `schema-parser.ts`) is **unchanged** — it only activates when explicitly opted in.

## Files Changed

- `src/schema-components-map.ts` — replaced unstable `sort()` calls in `enumsFirst()` and `discriminatorsFirst()` with the new stable `_stablePromoteToFront()` partition helper.

## Before / After

Given an OpenAPI spec with schemas declared in order `Alpha`, `Beta`, `Gamma` (where `Gamma` is an enum):

| | Before | After |
|---|---|---|
| `--sort-types` off | `Gamma`, `Alpha`, `Beta` (arbitrary) | `Gamma`, `Alpha`, `Beta` (enum first, rest in spec order) |
| `--sort-types` on | alphabetical | alphabetical (unchanged) |

### DIFF
--- a/.changeset/weak-llamas-promise.md
+++ b/.changeset/weak-llamas-promise.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
+fix: preserve OpenAPI spec property order when sortTypes is false (#151)
--- a/src/schema-components-map.ts
+++ b/src/schema-components-map.ts
@@ -108,4 +108,3 @@ export class SchemaComponentsMap {
     );
   }
 }
-
--- a/src/schema-components-map.ts
+++ b/src/schema-components-map.ts
@@ -69,21 +69,43 @@ export class SchemaComponentsMap {
     return this._data.find((c) => c.$ref === $ref) || null;
   };
 
-  // Ensure enums are at the top of components list
-  enumsFirst() {
-    this._data.sort((a, b) => {
-      if (Object.keys(a.rawTypeData || {}).includes("enum")) return -1;
-      if (Object.keys(b.rawTypeData || {}).includes("enum")) return 1;
-      return 0;
-    });
+  /**
+   * Partition-based stable reorder that moves items matching `predicate` to the
+   * front of `_data` while preserving the relative order of all other items.
+   *
+   * Unlike `Array.prototype.sort`, this never changes the relative order of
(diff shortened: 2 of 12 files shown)


smorimoto:
@cubic-dev-ai

PR #1722, opened 2026-04-24
fix(schema-parser): allOf intersection + nullable:true not generating | null union type

tenenger7125:
## Overview

First of all, I really enjoy using swagger-typescript-api in my project.
Thank you for maintaining such a great tool! 🙏

While using this library, I discovered a bug where `| null` was not being
appended to `allOf` intersection types even when `nullable: true` was specified.

---

## Problem

In OpenAPI 3.0, the correct pattern for a nullable `$ref` type is to wrap it
in `allOf` alongside `nullable: true`:

```json
"priorStudy": {
  "nullable": true,
  "allOf": [
    {
      "properties": {
        "patient": { ... }
      },
      "type": "object"
    },
    {
      "properties": {
        "studyIdInPacs": { "type": "string", "nullable": true },
        "category": {
          "allOf": [{ "$ref": "#/components/schemas/Category" }],
          "nullable": true
        }
      },
      "type": "object"
    }
  ]
}
```

**Expected output:**
```typescript
priorStudy:
  | ({
      patient: { ... };
    } & {
      studyIdInPacs: string | null;
      category: TCategory | null;
      ...
    })
  | null;  // ← should be here
```

**Actual output (before fix):**
```typescript
priorStudy: {
  patient: { ... };
} & {
  studyIdInPacs: string | null;
  category: TCategory | null;
  ...
};  // ← | null missing
```

---

## Root Cause

Two issues were found:

**1. `isNullMissingInType` in `schema-utils.ts`**

The function scanned the **entire multiline type string** for ` null`.
When inner fields contained `string | null` (e.g. `studyIdInPacs: string | null`),
it falsely concluded that `null` was already present at the top level,
and skipped adding `| null` to the intersection type.

```typescript
// Before (bug)
!type.includes(` ${this.config.Ts.Keyword.Null}`)  // ← matches inner fields too
```

**2. `$parsed` cache in `object.ts`**

Even after fixing `isNullMissingInType`, the `schema.$parsed` cache caused
the parser to skip re-evaluation on subsequent calls, so `safeAddNullToType`
was never reached for cached schemas.

---

## Fix

**`src/schema-parser/schema-utils.ts`**

Check only the **last line** of a multiline type string instead of the full string,
preventing false positives from inner field declarations.

```typescript
// After (fix)
const lastLine = type.trimEnd().split("\n").pop() ?? type;
return (
  !lastLine.includes(` ${nullKeyword}`) &&
  !lastLine.includes(`${nullKeyword} `)
);
```

**`src/schema-parser/base-schema-parsers/object.ts`**

Apply `safeAddNullToType` directly after `getInlineParseContent()` when
`nullable` is true, bypassing the `$parsed` cache issue.

```typescript
const rawFieldValue = this.schemaParserFabric
  .createSchemaParser({ schema: property, schemaPath: [...this.schemaPath, name] })
  .getInlineParseContent();

const fieldValue = nullable
  ? this.schemaUtils.safeAddNullToType(property, rawFieldValue)
  : rawFieldValue;
```

---

## Test

Added test cases to `tests/spec/nullable-3.0/schema.json`:

- `objectMaybeNullAllOf` — single `$ref` inside `allOf` + `nullable: true`
- `inlineObjectMaybeNullAllOf` — inline object inside `allOf` + `nullable: true`
- `intersectionMaybeNullAllOf` — **multiple objects in `allOf` + `nullable: true`**
  (the core bug case: inner fields contain `string | null`)

Ran `bun test:update` to update snapshots. Pre-existing snapshots that had
`nullable: true` but were missing `| null` were also corrected as a result
of this fix.

---

## Note

Pre-existing lint warnings exist in the codebase and are out of scope for this PR.

Two commits included:
- `fix`: source code changes (`schema-utils.ts`, `object.ts`)
- `test`: snapshot updates and new test cases

### DIFF
--- a/.changeset/heavy-moles-hammer.md
+++ b/.changeset/heavy-moles-hammer.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
+Add: null to allOf intersection types with nullable(true)
--- a/src/schema-parser/base-schema-parsers/object.ts
+++ b/src/schema-parser/base-schema-parsers/object.ts
@@ -45,12 +45,15 @@ export class ObjectSchemaParser extends MonoSchemaParser {
       const fieldName = this.typeNameFormatter.isValidName(name)
         ? name
         : this.config.Ts.StringValue(name);
-      const fieldValue = this.schemaParserFabric
+      const rawFieldValue = this.schemaParserFabric
         .createSchemaParser({
           schema: property,
           schemaPath: [...this.schemaPath, name],
         })
         .getInlineParseContent();
+      const fieldValue = nullable
+        ? this.schemaUtils.safeAddNullToType(property, rawFieldValue)
+        : rawFieldValue;
       const readOnly = (property as Record<string, unknown>).readOnly;
 
       const complexType = this.schemaUtils.getComplexType(property);
--- a/src/schema-parser/schema-utils.ts
+++ b/src/schema-parser/schema-utils.ts
@@ -149,15 +149,37 @@ export class SchemaUtils {
     return isRequired;
   };
 
+  // isNullMissingInType = (schema, type) => {
+  //   const { nullable, type: schemaType } = schema || {};
+  //   return (
+  //     (nullable ||
+  //       !!get(schema, "x-nullable") ||
+  //       schemaType === this.config.Ts.Keyword.Null) &&
+  //     typeof type === "string" &&
+  //     !type.includes(` ${this.config.Ts.Keyword.Null}`) &&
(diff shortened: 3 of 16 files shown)


js2me:
@tenenger7125 hello hello!!!   
THanks for your MR, LGTM!   

Please add `pnpm changeset` (`patch`)

tenenger7125:
> @tenenger7125 hello hello!!! THanks for your MR, LGTM!
> 
> Please add `pnpm changeset` (`patch`)

@js2me Thank you for the review! I've added the changeset as requested. 🙏

Issue #65, opened 2020-07-07
Underscores in interface names with specific schema definition.


Nalhin:
Spring automatically generates schema definitions with generic types (see the example for details). 
This causes the typescript interfaces to contain underscores in place of "«" symbols when such a schema is used.

Current result:
*Page_TemplateResponseDto_*

Expected result:
*PageTemplateResponseDto*

OA version - 2.0

Example schema:
```json
{
   "Page«TemplateResponseDto»":{
      "type":"object",
      "properties":{
         "content":{
            "type":"array",
            "items":{
               "$ref":"#/definitions/TemplateResponseDto"
            }
         },
         "empty":{
            "type":"boolean"
         },
         "first":{
            "type":"boolean"
         },
         "last":{
            "type":"boolean"
         },
         "number":{
            "type":"integer",
            "format":"int32"
         },
         "numberOfElements":{
            "type":"integer",
            "format":"int32"
         },
         "pageable":{
            "$ref":"#/definitions/Pageable"
         },
         "size":{
            "type":"integer",
            "format":"int32"
         },
         "sort":{
            "$ref":"#/definitions/Sort"
         },
         "totalElements":{
            "type":"integer",
            "format":"int64"
         },
         "totalPages":{
            "type":"integer",
            "format":"int32"
         }
      },
      "title":"Page«TemplateResponseDto»"
   }
}
```

js2me:
@Nalhin thanks for this issue!  
I'll fix that at next release (next day, GMT+3)

js2me:
It should be fixed in 1.11.0 release

PR #1726, opened 2026-04-24
fix(type-name-formatter): dedupe colliding identifiers via upfront precommit

### LINKED ISSUE #1724: bug: Duplicate export interface when two schemas normalize to same TS name
## Summary

Two OpenAPI schema keys that differ only by underscore/separator placement — e.g. `Foo_Bar` and `FooBar` — normalize to the same TypeScript identifier (`FooBar`) via [`src/type-name-formatter.js#L62-L66`](https://github.com/acacode/swagger-typescript-api/blob/13.0.8/src/type-name-formatter.js#L62-L66):

```js
const formattedName = _.replace(
  _.startCase(`${typePrefix}_${fixedModelName}_${typeSuffix}`),
  /\s/g,
  "",
);
```

`_.startCase` treats `_` as a word boundary, then the whitespace strip collapses the words, so `startCase("Foo_Bar")` and `startCase("FooBar")` both yield `"Foo Bar"`. Replacing `/\s/g` with `''` makes the final name to be `FooBar` in both cases.

`formattedModelNamesMap` is keyed by the raw input (`hashKey = prefix_name_suffix`), not by the formatted output, so the cache never detects collisions. The generator emits two `export interface FooBar { … }` declarations. TypeScript declaration-merges them, reporting `TS2717: Subsequent property declarations must have the same type` whenever the shapes differ (most commonly when both have a `type` discriminator with different const values).

## Versions

- `swagger-typescript-api@13.6.10` (latest at time of filing)
- `typescript@6.0.3`
- Node.js v24.11.0

## Minimal reproduction

`collision-repro.yaml`:

```yaml
openapi: 3.0.0
info: { title: repro, version: 1.0.0 }
paths:
  /a:
    get:
      responses:
        '200':
          content:
            application/json:
              schema: { $ref: '#/components/schemas/Foo_Bar' }
  /b:
    get:
      responses:
        '200':
          content:
            application/json:
              schema: { $ref: '#/components/schemas/FooBar' }
components:
  schemas:
    Foo_Bar:
      type: object
      required: [kind, legacyField]
      properties:
        kind: { type: string, enum: [legacy] }
        legacyField: { type: string }
    FooBar:
      type: object
      required: [kind, newField]
      properties:
        kind: { type: string, enum: [modern] }
        newField: { type: number }
```

Run:

```
npx -y swagger-typescript-api@latest generate -p ./collision-repro.yaml -o ./out
```

Generated `./out/Api.ts` contains:

```ts
export interface FooBar {
  kind: "legacy";
  legacyField: string;
}

export interface FooBar {
  kind: "modern";
  newField: number;
}
```

### Showing the TS error

The generator writes `// @ts-nocheck` at the top of the output file, which silences the diagnostic inside that file. The bug is still present — any downstream consumer that imports `FooBar` observes the declaration-merged type — but to surface the error directly, strip the pragma and run `tsc`:

```
sed -i '' 's|// @ts-nocheck||' ./out/Api.ts
npx -y -p typescript@latest tsc --strict --noEmit ./out/Api.ts
```

Output:

```
out/Api.ts(19,3): error TS2717: Subsequent property declarations must have the same type.
  Property 'kind' must be of type '"legacy"', but here has type '"modern"'.
```

## Context

Our schema is produced by an OpenAPI generator for another language (Scala), which independently emits both underscore-wrapped names (for parameterized discriminator variants) and plain names (for newly-added flat schemas). Having both shapes live side by side in a single spec is idiomatic output from that generator, not a schema-authoring mistake, so we can't ask the upstream producer to rename things.

## Expected

Two distinct TypeScript identifiers — for example `FooBar` and `FooBar1`, or any deterministic disambiguation. The existing `onFormatTypeName` hook provides a per-name workaround, but the default output for a valid spec should not require a hook.

## Actual

Two `export interface FooBar` declarations in the generated file. TypeScript declaration-merges them and errors on any conflicting field.

## Edge cases a fix should handle

- **Pre-existing suffixed schema**: the raw spec may already contain `FooBar1` as a legitimate schema. A naive `baseName + "1"` fallback would collide again. The fix needs to pick the smallest integer suffix not already used by any raw-name-derived identifier in the same run — i.e. check the full set of formatted names already committed, not just the immediate pair.
- **Order sensitivity**: schemas are processed in source order; the first-seen name keeps the unsuffixed identifier. Consistent with other order-dependent parts of the generator, but worth documenting.
- **Hook interaction**: `onFormatTypeName` should continue to see the pre-disambiguation formatted name; collision suffixing should run on the hook's return value so that two hooks returning identical strings are also disambiguated.

## Related

[#65](https://github.com/acacode/swagger-typescript-api/issues/65) introduced the `startCase` normalization that underpins this. The normalization is correct for its original case (generic-type wrappers like `Page«TemplateResponseDto»`); the missing piece is collision detection on the **formatted** output.

## Possible fix

Track formatted names in a second map keyed by the formatted output; on a collision from a distinct raw input, append the smallest integer suffix not already present in that map. Happy to put up a PR — let me know if that's welcome.

mlewando-cp:
## Summary

Two OpenAPI schema keys that differ only in separator placement — e.g. `Foo_Bar` and `FooBar` — both normalize to the same TypeScript identifier (`FooBar`) via [`src/type-name-formatter.ts`](https://github.com/acacode/swagger-typescript-api/blob/main/src/type-name-formatter.ts)'s `startCase` + whitespace-strip pipeline. The previous cache is keyed by the raw input (`hashKey = prefix_name_suffix`), not by the formatted output, so collisions go undetected and the generator emits two `export interface FooBar { … }` declarations — TypeScript declaration-merges them and reports `TS2717` whenever the shapes differ.

This PR resolves collisions **upfront**, in a new `precommit(rawNames)` pass the generator invokes once after loading schema components and before schema parsing begins. `format()` shrinks to a pure cache lookup plus a no-collision-logic fallback for dynamically-discovered names (enum keys, `extractEnums` / `extractResponses` results). All formatting logic is concentrated in a single private `computeFormattedName()` helper so `format()`, `precommit()`, and the new `disableFormatTypeNames` / `typeNameSeparator` options share one pipeline.

Fixes #1724.

## Why not resolve collisions in `format()`?

An earlier iteration of this PR tried a mid-flight "evict-and-rename" approach inside `format()`: when a canonical incoming name collides with a non-canonical incumbent, move the incumbent to a suffixed slot so the user-declared name survives. It **corrupts downstream output**.

The schema parser calls `format()` during Phase 1 (from `schema-utils.ts`, `schema-parser.ts`, `schema-routes.ts`) and **captures the return value directly into inline type strings** — e.g. `this.request<FooBar, any>` for a route response. Later Phase-2 work (`prepareModelType` → `collectModelTypes`) calls `format()` again per schema and captures the result into `preparedModelType.name`. Phase 2 happens after Phase 1, so definitions always see the final cache state and come out consistent. But any eviction that renames a raw schema mid-flight leaves the earlier Phase-1 captures frozen at the old name, and a route that returns `Foo_Bar` can end up generic-parameterized on `FooBar` — which is now the shape of an entirely different schema. A route-reference regression test in this PR would have caught it.

Pre-committing every schema name before the parser runs sidesteps the problem entirely: every `format()` call returns the final name on the first shot.

### Two-pass resolution in `precommit`

1. **Pass 1 — canonicals claim their slots.** For every raw name where `rawName === computeFormattedName(rawName)` (i.e. the name already matches its formatted output), reserve that identifier. User-declared names like `FooBar`, `FooBar1`, `User`, `V1Response` all survive regardless of source order.
2. **Pass 2 — non-canonicals suffix-until-free.** For every remaining raw name (e.g. `Foo_Bar` → `FooBar`), append the smallest integer suffix not already used by any pre-committed canonical or earlier non-canonical, producing deterministic `FooBar`, `FooBar1`, `FooBar2`, …

### `format()` becomes a lookup

```ts
format = (name, options = {}) => {
  // ... typePrefix/typeSuffix/hashKey setup ...
  const cached = this.formattedModelNamesMap.get(hashKey);
  if (cached !== undefined) return cached;

  // Fallback (no collision check): compute inline, cache, return.
  // Reached only for enum keys and schemas added after precommit.
  const result = this.computeFormattedName(name, schemaType, typePrefix, typeSuffix);
  this.formattedModelNamesMap.set(hashKey, result);
  return result;
};
```

### Single source of truth for the pipeline

`computeFormattedName` is the only place startCase, `disableFormatTypeNames`, `typeNameSeparator`, the ALL_CAPS fast path, and the `onFormatTypeName` hook are applied. `precommit` and the `format()` fallback both go through it, so all three formatting options (this PR's collision resolution + `disableFormatTypeNames` + `typeNameSeparator`) compose without special-casing.

### Enum keys are intentionally excluded

Enum keys live in their own per-enum scope (`EnumA.Bar` and `EnumB.Bar` are both valid TS). Cross-enum collisions don't need disambiguation, so `precommit` and `usedFormattedTypeNames` only track `schemaType === "type-name"`. Enum keys always hit the fallback path in `format()`, which preserves existing behavior.

### Hook interaction

Collision detection runs on the **hook output**, not the raw `startCase` output. If `onFormatTypeName` maps two distinct raw names to the same string, one still ends up suffixed.

## Minimal repro (pre-fix)

```yaml
# collision-repro.yaml
openapi: 3.0.0
info: { title: repro, version: 1.0.0 }
paths:
  /a:
    get:
      responses:
        '200':
          content:
            application/json:
              schema: { $ref: '#/components/schemas/Foo_Bar' }
  /b:
    get:
      responses:
        '200':
          content:
            application/json:
              schema: { $ref: '#/components/schemas/FooBar' }
components:
  schemas:
    Foo_Bar:
      type: object
      required: [kind, legacyField]
      properties:
        kind: { type: string, enum: [legacy] }
        legacyField: { type: string }
    FooBar:
      type: object
      required: [kind, newField]
      properties:
        kind: { type: string, enum: [modern] }
        newField: { type: number }
```

Before: two `export interface FooBar` declarations, TS2717 once `@ts-nocheck` is stripped. After: `export interface FooBar` (for raw `FooBar`, with `kind: "modern"`) and `export interface FooBar1` (for raw `Foo_Bar`, with `kind: "legacy"`); route references are consistent.

## Tests

Seven new cases under `tests/spec/name-collision/`:

| Case | What it asserts |
|---|---|
| Basic collision | `Foo_Bar` + `FooBar` → `FooBar` + `FooBar1`; both `kind` literals preserved |
| Pre-existing suffix, natural order | `Foo_Bar`, `FooBar`, `FooBar1` in that order → `FooBar`, `FooBar1`, `FooBar2`; no `FooBar11` |
| Pre-existing suffix, reversed order | Same three schemas reversed → same outcome (order-independent) |
| Multiple canonicals | `FooBar`, `Foo_Bar`, `FooBar1`, `FooBar2` → non-canonical lands at `FooBar3`, canonicals preserved |
| Hook interaction | `onFormatTypeName: () => "Shared"` → `Shared` + `Shared1` |
| Repeat reference | `Dog` referenced from three endpoints → exactly one `export interface Dog` |
| **Route-reference correctness** | With `generateClient: true`, each route's generic parameter matches the corresponding definition by its unique property. Catches the eviction-era regression described above. |

## Context

Our client is generated from a Scala backend's OpenAPI spec. Scala's generator naturally emits both underscore-wrapped names (parameterized discriminator variants) and plain names (flat schemas), so having `Foo_Bar` and `FooBar` side-by-side is idiomatic output, not a schema authoring mistake. The numeric suffix convention (`Foo_Bar1` in some specs) comes from the Scala generator itself, not from `swagger-typescript-api` — the suffix loop in this PR skips past it correctly.

Historical context: [#65](https://github.com/acacode/swagger-typescript-api/issues/65) introduced the `startCase` normalization that underpins the collision. This PR doesn't change that normalization; it adds collision-aware resolution on top of it.

## Test plan

- [x] `bun run test` → 198 passing (189 pre-existing + 7 name-collision + 2 for the new `disableFormatTypeNames`/`typeNameSeparator` suites still green on top of this PR)
- [x] `bun run build` clean
- [x] `bun run format:check` clean
- [x] `bunx biome check` on modified files clean
- [x] Manual route-reference smoke test on a client-generating spec: route generics match definitions

### DIFF
--- a/.changeset/quiet-otters-resolve.md
+++ b/.changeset/quiet-otters-resolve.md
@@ -1,0 +1,27 @@
+---
+"swagger-typescript-api": patch
+---
+
+Dedupe colliding TypeScript identifiers produced by the `TypeNameFormatter`.
+
+Two OpenAPI schema keys that differ only in separator placement — e.g.
+`Foo_Bar` and `FooBar` — used to collapse to the same identifier via
+`startCase` + whitespace-strip and emit two `export interface FooBar`
+declarations (TS2717 whenever the shapes differed).
+
+`TypeNameFormatter` now exposes a `precommit(rawNames)` method the generator
+calls once after loading schema components and before schema parsing. It
+resolves every raw name in two passes — canonical names (raw === formatted
+output) claim their slot first, then non-canonical names suffix-until-free —
+so user-declared identifiers like `FooBar1` are preserved regardless of
+source order, and collisions deterministically produce `FooBar`, `FooBar1`,
+`FooBar2`, … References to each schema (including inline generics in route
+handlers) stay consistent with the emitted `export interface` declarations.
+
+`format()` is now a pure cache lookup with a fallback for names discovered
+after precommit (enum keys, `extractEnums`/`extractResponses` results). All
+formatting logic is concentrated in a single private `computeFormattedName`
+helper, so the new behavior composes cleanly with `disableFormatTypeNames`
+and `typeNameSeparator`.
+
+Fixes #1724.
--- a/src/code-gen-process.ts
+++ b/src/code-gen-process.ts
@@ -140,6 +140,13 @@ export class CodeGenProcess {
         compact(["schemas", this.config.extractResponses && "responses"]),
       );
 
+    // Resolve the TypeScript identifier for every schema component upfront,
+    // before the parser starts calling `format()`. This lets `format()` stay
+    // a pure cache lookup and keeps collision handling concentrated in one
+    // place so results are source-order independent and inline type strings
+    // captured during schema parsing can't go stale. See #1724.
+    this.typeNameFormatter.precommit(componentsToParse.map((c) => c.typeName));
+
     const parsedSchemas = componentsToParse.map((schemaComponent) => {
       const parsed = this.schemaParserFabric.parseSchema(
         schemaComponent.rawTypeData,
--- a/src/type-name-formatter.ts
+++ b/src/type-name-formatter.ts
@@ -7,70 +7,119 @@ type FormattingSchemaType = "enum-key" | "type-name";
 
 export class TypeNameFormatter {
   formattedModelNamesMap = new Map<string, string>();
+  // Tracks formatted output → raw name for top-level type names. Populated
+  // by `precommit` during the resolution pass so `format()` can stay a pure
+  // lookup. Enum keys are NOT tracked here — they live inside enum blocks
+  // (`EnumA.Bar` and `EnumB.Bar` are both valid), so cross-enum collisions
+  // don't need to be disambiguated. See issue #1724.
+  usedFormattedTypeNames = new Map<string, string>();
   config: CodeGenConfig;
 
   constructor(config: CodeGenConfig) {
     this.config = config;
   }
 
+  /**
+   * Return the TypeScript identifier for a raw OpenAPI name. Fast path is a
+   * cache hit on names resolved by `precommit`. The fallback (for names
+   * discovered after precommit, e.g. enum keys inside schemas or types added
+   * dynamically by `extractEnums`/`extractResponses`) computes the identifier
+   * inline WITHOUT collision handling — collision resolution is the sole
+   * responsibility of `precommit`. Callers expecting collision safety must
+   * list every raw name in the precommit input.
+   */
   format = (name: string, options: { type?: FormattingSchemaType } = {}) => {
     const schemaType = options.type ?? "type-name";
-
-    let typePrefix: string;
-    let typeSuffix: string;
-
-    if (schemaType === "enum-key") {
-      typePrefix = this.config.enumKeyPrefix;
-      typeSuffix = this.config.enumKeySuffix;
-    } else {
-      typePrefix = this.config.typePrefix;
-      typeSuffix = this.config.typeSuffix;
-    }
-
-    const typeNameSeparator = this.config.typeNameSeparator;
-
+    const { typePrefix, typeSuffix } = this.getAffixes(schemaType);
     const hashKey = `${typePrefix}_${name}_${typeSuffix}`;
 
-    if (this.formattedModelNamesMap.has(hashKey)) {
-      return this.formattedModelNamesMap.get(hashKey);
-    }
+    const cached = this.formattedModelNamesMap.get(hashKey);
+    if (cached !== undefined) return cached;
 
     if (typeof name !== "string") {
       consola.warn("wrong model name", name);
       return name;
     }
 
-    let resultName = name;
+    // Fallback: compute inline, no collision check.
+    const result = this.computeFormattedName(
+      name,
+      schemaType,
+      typePrefix,
+      typeSuffix,
+    );
+    this.formattedModelNamesMap.set(hashKey, result);
+    return result;
+  };
 
-    if (this.config.disableFormatTypeNames) {
-      resultName = compact([typePrefix, resultName, typeSuffix]).join(
-        typeNameSeparator,
-      );
-    } else {
-      // https://github.com/acacode/swagger-typescript-api/issues/1260
-      if (/^(?!\d)([A-Z0-9_]{1,})$/g.test(resultName)) {
-        resultName = compact([typePrefix, resultName, typeSuffix]).join(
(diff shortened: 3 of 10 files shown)


js2me:
@mlewando-cp Thank you!

PR #1754, opened 2026-05-19
Replace generateUnionEnums with unified enumStyle option

### LINKED ISSUE #1040: Generate objects instead of enums

Upgrade220:
Closes https://github.com/acacode/swagger-typescript-api/issues/1040

## Problem

The library currently exposes `generateUnionEnums` as a boolean flag to switch enum output to union types. PR #1281 proposes adding a second boolean `generateConstObjectEnums` for `as const` object output. I think adding a second independent boolean is the wrong direction — two mutually exclusive flags that silently override each other is confusing API design, and it still leaves a gap: `ContentType` in the http-client is always emitted as a traditional `enum` regardless of what the user configures, making the output inconsistent.

## Proposal

I propose replacing both flags with a single unified `enumStyle` option:

```typescript
enumStyle?: "enum" | "union" | "const"
// default: "enum"
```

This makes the three output modes explicit and mutually exclusive by construction. `generateUnionEnums` is preserved as a deprecated alias — it still works and emits a runtime warning pointing users to the new option, so nothing breaks for existing consumers.

I also propose extending the `"const"` style to cover `ContentType` in the http-client templates. When `enumStyle: "const"` is set, `ContentType` is emitted as:

```typescript
export const ContentType = {
  Json: "application/json",
  ...
} as const;
export type ContentType = (typeof ContentType)[keyof typeof ContentType];
```

This makes the entire generated output consistent — schema enums and internal http-client constants follow the same pattern.

## Changes

- New `enumStyle: "enum" | "union" | "const"` config option and `--enum-style` CLI flag
- `generateUnionEnums: true` auto-promotes to `enumStyle: "union"` with a deprecation warning
- `"const"` style generates `as const` objects with a companion type alias for all schema enums
- `"const"` style also converts `ContentType` in both Axios and Fetch http-client templates
- `enum-data-contract.ejs` subtemplate updated to support all three styles
- `Ts.Keyword.Const` added for template authors who use `codeGenConstructs`
- Snapshot test added for `enumStyle: "const"`

---
## Summary by cubic
Replaces `generateUnionEnums` with a unified `enumStyle` option ("enum" | "union" | "const") to control enum output, and applies the `"const"` style to `ContentType` in Axios and Fetch clients. Adds a CLI flag, deprecates the old option with a warning, keeps the default as `"enum"`, and fixes an internal export leak for `const` enums.

- **New Features**
  - Added `enumStyle` ("enum" | "union" | "const") and `--enum-style` CLI flag; `generateUnionEnums` remains as a deprecated alias that auto-promotes to `enumStyle: "union"` with a warning.
  - `"const"` style emits `as const` objects with a companion type alias and updates `ContentType` in both clients; parsers/discriminator handling/templates updated; snapshot tests added.

- **Bug Fixes**
  - `const` handler now respects `internal` and does not export the type alias for internal contracts.

<sup>Written for commit cb32443ab8f8290d66f221f729989bede155c8b3. Summary will update on new commits. <a href="https://cubic.dev/pr/acacode/swagger-typescript-api/pull/1754?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/add-enum-style-option.md
+++ b/.changeset/add-enum-style-option.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": minor
+---
+
+Add `enumStyle` option ("enum" | "union" | "const") to control enum output format. `"const"` generates `as const` objects with a companion type alias, including the built-in `ContentType` in the http-client. `generateUnionEnums` is deprecated in favor of `enumStyle: "union"`.
--- a/index.ts
+++ b/index.ts
@@ -183,8 +183,13 @@ const generateCommand = defineCommand({
     },
     "generate-union-enums": {
       type: "boolean",
-      description: 'generate all "enum" types as union types (T1 | T2 | TN)',
+      description: '(deprecated) generate all "enum" types as union types — use --enum-style=union instead',
       default: codeGenBaseConfig.generateUnionEnums,
+    },
+    "enum-style": {
+      type: "string",
+      description: 'enum output style: "enum" (default), "union" (T1 | T2 | TN), or "const" (as const object + type alias)',
+      default: codeGenBaseConfig.enumStyle,
     },
     "http-client": {
       type: "string",
@@ -323,6 +328,7 @@ const generateCommand = defineCommand({
       generateResponses: args.responses,
       generateRouteTypes: args["route-types"],
       generateUnionEnums: args["generate-union-enums"],
+      enumStyle: args["enum-style"] as "enum" | "union" | "const" | undefined,
       httpClientType:
         args["http-client"] || args.axios
           ? HTTP_CLIENT.AXIOS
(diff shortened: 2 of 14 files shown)


js2me:
@Upgrade220 Thank you or Спасибо большое за вклад  за вклад в инструмент :)

Upgrade220:
> @Upgrade220 Thank you or Спасибо большое за вклад за вклад в инструмент :)

Да всегда пожалуйста) Я давненько уже пользуюсь swagger-typescript-api и считаю данный инструмент одним из передовых в данной теме для TS)

Скорее всего скоро принесу еще пачечку МРов с другими изменениями

PR #1756, opened 2026-05-19
Fix: query params object incorrectly required when all fields are optional (extractRequestParams)

### LINKED ISSUE #1755: extractRequestParams: query params object required even when all fields are optional
## Describe the bug

When using `extractRequestParams: true`, if a route has only optional query parameters and no path parameters, the generated method signature incorrectly marks the combined params object as required:

```typescript
// generated — wrong
getItems(query: GetItemsParams, params?: RequestParams): Promise<...>
```

The caller is forced to pass an explicit argument even if they want to use defaults:
```typescript
// must write this
api.items.getItems({})

// but should be able to write this
api.items.getItems()
```

## Expected behavior

When all query fields are optional and there are no path params, the combined params object should have a default value of `{}`:

```typescript
// expected
getItems(query: GetItemsParams = {}, params?: RequestParams): Promise<...>
```

## Root cause

`procedure-call.ejs` hardcodes `optional: false` for the `requestParams` argument regardless of whether all query fields are optional. The `allFieldsAreOptional` flag is already computed correctly in the source (`requestParamsSchema.typeData.allFieldsAreOptional`) but is never surfaced to the template.

## Steps to reproduce

1. Define an OpenAPI route with only optional query parameters and no path parameters
2. Run generation with `--extract-request-params`
3. Observe the generated method signature — the params object is required

Upgrade220:
Closes https://github.com/acacode/swagger-typescript-api/issues/1755

## Problem

When using `extractRequestParams: true`, if a route has only optional query parameters and no path parameters, the generated method signature forces callers to pass an explicit argument even though every field inside it is optional:

```typescript
// generated — wrong
getItems(query: GetItemsParams, params?: RequestParams)

// caller is forced to write
api.items.getItems({})
// instead of simply
api.items.getItems()
```

The root cause is that `procedure-call.ejs` hardcodes `optional: false` for the combined params argument. The `allFieldsAreOptional` flag is already computed correctly inside `requestParamsSchema.typeData` by the existing `ObjectSchemaParser`, but it was never surfaced to the template.

## Solution

I propose a minimal source-level fix: expose `requestParamsSchema?.typeData?.allFieldsAreOptional` as `requestParamsOptional` on the route's request info object in `parseRouteInfo()` (`schema-routes.ts`). The template then reads this flag instead of hardcoding `false`, and adds `= {}` as the default value when the params object is optional.

No schema construction changes are needed — the data was already correct, just unexposed.

```typescript
// generated — correct
getItems(query: GetItemsParams = {}, params?: RequestParams)
```

## Changes

- `src/schema-routes/schema-routes.ts` — add `requestParamsOptional` to the route request info
- `types/index.ts` — add `requestParamsOptional?: boolean` to `ParsedRouteRequest`
- `templates/default/procedure-call.ejs` — read `route.request.requestParamsOptional` instead of hardcoded `false`
- `templates/modular/procedure-call.ejs` — same change
- `tests/spec/extractRequestParams/schema.json` — add a route with only optional query params to cover this case
- Snapshot updates across extended tests where routes with all-optional query params now correctly get `= {}`

---
## Summary by cubic
Fixes `extractRequestParams` so routes with only optional query params no longer force callers to pass `{}`. Generated methods now make the combined query object optional and default it to `{}`.

- **Bug Fixes**
  - Mark the combined query params object as optional and add `= {}` when all its fields are optional and there are no path params.
  - Surface `requestParamsSchema.typeData.allFieldsAreOptional` as `requestParamsOptional` in `schema-routes.ts`, and add `requestParamsOptional` to `ParsedRouteRequest` in `types/index.ts`.
  - Update `templates/*/procedure-call.ejs` to respect `requestParamsOptional` and set `defaultValue: '{}'` when applicable.
  - Add a covering test route and update snapshots; add a patch changeset.

<sup>Written for commit 209cb7ae53d78b992ca9db32edea3ccb28a77fc7. Summary will update on new commits. <a href="https://cubic.dev/pr/acacode/swagger-typescript-api/pull/1756?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/fix-optional-query-params.md
+++ b/.changeset/fix-optional-query-params.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
+Fix: combined query params object now correctly gets a default value of `{}` when all its fields are optional and no path params are present (extractRequestParams mode)
--- a/src/schema-routes/schema-routes.ts
+++ b/src/schema-routes/schema-routes.ts
@@ -1326,6 +1326,8 @@ export class SchemaRoutes {
         security: hasSecurity,
         method: method,
         requestParams: requestParamsSchema,
+        requestParamsOptional:
+          requestParamsSchema?.typeData?.allFieldsAreOptional ?? false,
 
         payload: specificArgs.body,
         query: specificArgs.query,
--- a/templates/default/procedure-call.ejs
+++ b/templates/default/procedure-call.ejs
@@ -31,8 +31,9 @@ const rawWrapperArgs = config.extractRequestParams ?
     _.compact([
         requestParams && {
           name: extractedRequestParamsName,
-          optional: false,
+          optional: route.request.requestParamsOptional,
           type: getInlineParseContent(requestParams),
+          defaultValue: route.request.requestParamsOptional ? '{}' : undefined,
         },
         ...(!requestParams ? pathParams : []),
         payload,
--- a/templates/modular/procedure-call.ejs
+++ b/templates/modular/procedure-call.ejs
@@ -31,8 +31,9 @@ const rawWrapperArgs = config.extractRequestParams ?
     _.compact([
         requestParams && {
           name: extractedRequestParamsName,
-          optional: false,
+          optional: route.request.requestParamsOptional,
           type: getInlineParseContent(requestParams),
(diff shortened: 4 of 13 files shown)


js2me:
@Upgrade220 thank you!!!!

PR #1762, opened 2026-05-21
Fix ContentType ignoring enumStyle: "union" in http-client

### LINKED ISSUE #1761: ContentType in http-client ignores enumStyle: "union"
When `enumStyle: "union"` is set, all user-defined enums are correctly generated as plain type aliases (e.g. `export type Status = "active" | "inactive"`). However, the built-in `ContentType` in the generated http-client is still emitted as a regular TypeScript `enum` regardless of the setting.

Additionally, the generated API call sites use `ContentType.Json`, `ContentType.FormData`, etc. — dot-property access that requires a runtime object. Since a plain type alias has no runtime representation, this produces invalid TypeScript under `isolatedModules: true` and breaks at runtime.

This was missed in #1754 where `enumStyle` was introduced. The `const` branch was handled for ContentType (it generates a `const` object + type alias), but no `union` branch was added.

**Expected behaviour:**

With `enumStyle: "union"`:
```ts
// http-client
export type ContentType = "application/json" | "application/vnd.api+json" | "multipart/form-data" | "application/x-www-form-urlencoded" | "text/plain";

// generated API call
createPet(data: Pet, params: RequestParams = {}) =>
  this.request<Pet, any>({
    path: `/pets`,
    method: "POST",
    body: data,
    type: "application/json",  // ← string literal, not ContentType.Json
    ...params,
  })
```

**Actual behaviour:**

`ContentType` is generated as `export enum ContentType { ... }` and call sites use `type: ContentType.Json`.

Apologies for the oversight — this should have been caught and included in #1754.

Upgrade220:
## Problem

When `enumStyle: "union"` is configured, user-defined enums are generated as plain type aliases. The built-in `ContentType` in the http-client was not covered by this path — it fell through to the `else` branch and was always generated as a regular `enum`. This was overlooked in #1754 when `enumStyle` was introduced; only the `const` case was handled for ContentType.

Beyond the declaration, all call sites inside the http-client class body (`contentFormatters`, `type || ContentType.Json`, `type !== ContentType.FormData`) and in generated API methods (`type: ContentType.Json`) used dot-property access, which is invalid when ContentType is a type alias with no runtime object. This also causes `isolatedModules` errors in the modular output's import.

Fixes #1761.

## Solution

- Adds a `union` branch to both `fetch-http-client.ejs` and `axios-http-client.ejs` that emits `export type ContentType = "application/json" | ...`
- Introduces `templates/base/content-type-accessors.ejs`, a shared partial (following the `route-docs.ejs` pattern) that returns the correct accessor string for each content type based on `enumStyle`. It serves two consumers with different key conventions: http-client templates use camelCase keys (`CT.Json`), procedure-call templates use UPPER_SNAKE keys (`requestContentKind["JSON"]`) matching the `CONTENT_KIND` constants from `schema-routes.ts`
- Updates `modular/api.ejs` to emit `import { …, type ContentType, … }` (inline type import) in union mode, avoiding `isolatedModules` errors
- Adds a `union-enums-http-client` test case to the existing `enumStyle` spec

---
## Summary by cubic
Makes `ContentType` in generated http-clients respect `enumStyle: "union"`, fixing invalid property access and `isolatedModules` issues. Unifies content-type access across templates to work with both enum and union modes.

- **Bug Fixes**
  - Generate `ContentType` as a union type in `fetch-http-client.ejs` and `axios-http-client.ejs` when `enumStyle` is `"union"`.
  - Use shared accessors via `templates/base/content-type-accessors.ejs` in http-clients and procedure-call templates to avoid dot-property access in union mode.
  - In modular builds, import `type ContentType` from the http-client when in union mode to satisfy `isolatedModules`.
  - Add `union-enums-http-client` test coverage.

- **Migration**
  - If you reference `ContentType.*` at runtime while using `enumStyle: "union"`, switch to string literals (e.g., `"application/json"`) or treat `ContentType` as a type-only import. No changes needed for generated API usage.

<sup>Written for commit 90cf9cc9d522d043a602b651a1fa48a365427124. Summary will update on new commits. <a href="https://cubic.dev/pr/acacode/swagger-typescript-api/pull/1762?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/.changeset/fix-union-content-type.md
+++ b/.changeset/fix-union-content-type.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
+Fix `ContentType` in http-client not respecting `enumStyle: "union"`. It now generates a plain type alias instead of an enum, and all call sites emit string literals instead of `ContentType.Json` etc.
--- a/templates/base/content-type-accessors.ejs
+++ b/templates/base/content-type-accessors.ejs
@@ -1,0 +1,18 @@
+<%
+const { config } = it;
+const isUnion = config.enumStyle === "union";
+
+// camelCase keys (Json, FormData, …) are used by http-client templates as CT.Json, CT.FormData, …
+// They mirror the ContentType property names so the access pattern is identical regardless of enumStyle.
+const v = {
+  Json:       isUnion ? '"application/json"'                  : "ContentType.Json",
+  JsonApi:    isUnion ? '"application/vnd.api+json"'          : "ContentType.JsonApi",
+  FormData:   isUnion ? '"multipart/form-data"'               : "ContentType.FormData",
+  UrlEncoded: isUnion ? '"application/x-www-form-urlencoded"' : "ContentType.UrlEncoded",
+  Text:       isUnion ? '"text/plain"'                        : "ContentType.Text",
+};
+
+// UPPER_SNAKE aliases (JSON, FORM_DATA, …) are used by procedure-call templates as
+// requestContentKind["JSON"], matching the CONTENT_KIND constants produced by schema-routes.ts.
+return { ...v, JSON: v.Json, JSON_API: v.JsonApi, FORM_DATA: v.FormData, URL_ENCODED: v.UrlEncoded, TEXT: v.Text };
+%>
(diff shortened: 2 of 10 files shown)


js2me:
LGTM, thanks!

PR #1803, opened 2026-06-29
Handle object-shaped not schemas without crashing

### LINKED ISSUE #1798: `swagger-typescript-api` crashes on OpenAPI `not: {}` in `13.12.2`, while `13.0.28` succeeds
### Describe the bug

`swagger-typescript-api` crashes when the OpenAPI schema contains `not: {}`.

This works in `13.0.28`, but fails in `13.12.2`.

### Minimal repro

Repro spec:
```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "swagger-typescript-api not repro",
    "version": "1.0.0"
  },
  "paths": {},
  "components": {
    "schemas": {
      "RelatedProductVariant": {
        "type": "object",
        "properties": {
          "variantId": {
            "type": "integer"
          },
          "variantReferenceKey": {
            "type": "string"
          },
          "isMainVariant": {
            "type": "boolean"
          }
        }
      },
      "ProductVariant": {
        "type": "object",
        "properties": {
          "relatedVariants": {
            "oneOf": [
              {
                "type": "array",
                "minItems": 1,
                "items": {
                  "allOf": [
                    {
                      "$ref": "#/components/schemas/RelatedProductVariant"
                    },
                    {
                      "type": "object",
                      "required": [
                        "variantId",
                        "isMainVariant"
                      ],
                      "properties": {
                        "variantId": {
                          "type": "integer"
                        },
                        "variantReferenceKey": {
                          "not": {}
                        }
                      }
                    }
                  ]
                }
              },
              {
                "type": "array",
                "minItems": 1,
                "items": {
                  "allOf": [
                    {
                      "$ref": "#/components/schemas/RelatedProductVariant"
                    },
                    {
                      "type": "object",
                      "required": [
                        "variantReferenceKey",
                        "isMainVariant"
                      ],
                      "properties": {
                        "variantReferenceKey": {
                          "type": "string"
                        },
                        "variantId": {
                          "not": {}
                        }
                      }
                    }
                  ]
                }
              }
            ]
          }
        }
      }
    }
  }
}
```

Command:

```bash
swagger-typescript-api generate \
  --axios \
  -p ./docs/swagger-typescript-api-not-repro.json \
  -o ./tmp/generated \
  -n api.ts
```

### Current behavior

`13.12.2` throws:

```text
TypeError: (this.schema[complexType] || []).map is not a function
    at ComplexSchemaParser.parse (...)
```

### Expected behavior

The generator should not crash on `not: {}`.

If this schema shape is unsupported, a clear validation error would still be much better than a runtime exception.

### Version info

Fails:

- `swagger-typescript-api@13.12.2`

Works:

- `swagger-typescript-api@13.0.28`

### Suspected cause

It looks like `ComplexSchemaParser` handles `not` as a complex schema, but later assumes the complex schema value is array-like when building the description.

`13.0.28` uses Lodash:

```ts
lodash.map(this.schema[complexType], "description")
```

That safely returns `[]` for `not: {}`.

`13.12.2` uses native `.map(...)`:

```ts
(this.schema[complexType] || []).map((item) => item?.description)
```

For `not: {}`, that becomes `{}` and throws because `{}` does not have `.map`.

### Additional context

In a larger real-world spec, removing only the two `not: {}` entries made generation succeed, which seems to confirm this is the direct trigger.

### Relevant source

- `13.0.28`: https://raw.githubusercontent.com/acacode/swagger-typescript-api/13.0.28/src/schema-parser/base-schema-parsers/complex.ts
- `13.12.2`: https://raw.githubusercontent.com/acacode/swagger-typescript-api/v13.12.2/src/schema-parser/base-schema-parsers/complex.ts

morgan-coded:
A `not: {}` schema could reach the complex schema parser and crash because the parser expected complex keyword values to be arrays.

This adds array guards for complex keyword parsing, so non-array values take the same empty-result path as before while array-valued `oneOf`, `anyOf`, and `allOf` behavior stays unchanged. I checked the issue repro against the failing path, then ran install, format, build, and the full test suite with 275 tests passing.

Fixes #1798

### DIFF
--- a/.changeset/silly-waves-learn.md
+++ b/.changeset/silly-waves-learn.md
@@ -1,0 +1,5 @@
+---
+"swagger-typescript-api": patch
+---
+
+Handle object-shaped not schemas without crashing
--- a/src/schema-parser/base-schema-parsers/complex.ts
+++ b/src/schema-parser/base-schema-parsers/complex.ts
@@ -5,6 +5,11 @@ import { MonoSchemaParser } from "../mono-schema-parser.js";
 export class ComplexSchemaParser extends MonoSchemaParser {
   override parse() {
     const complexType = this.schemaUtils.getComplexType(this.schema);
+    const complexSchemaItems: Array<Record<string, unknown>> = Array.isArray(
+      this.schema[complexType],
+    )
+      ? this.schema[complexType]
+      : [];
     const simpleSchema = omit(
       this.schema,
       Object.keys(this.schemaParser._complexSchemaParsers),
@@ -23,11 +28,7 @@ export class ComplexSchemaParser extends MonoSchemaParser {
       name: this.typeName,
       description: this.schemaFormatters.formatDescription(
         this.schema.description ||
-          compact(
-            (this.schema[complexType] || []).map(
-              (item: any) => item?.description,
-            ),
-          )[0] ||
+          compact(complexSchemaItems.map((item) => item.description))[0] ||
           "",
       ),
       content:
--- a/src/schema-parser/base-schema-parsers/object.ts
+++ b/src/schema-parser/base-schema-parsers/object.ts
@@ -58,6 +58,16 @@ export class ObjectSchemaParser extends MonoSchemaParser {
 
       const complexType = this.schemaUtils.getComplexType(property);
       const rawDataComplexType = this.schemaUtils.getComplexType(rawTypeData);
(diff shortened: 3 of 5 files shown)


js2me:
@morgan-coded hello hello!!!!   
Please add changeset `pnpm changeset`

js2me:
@morgan-coded thank you for your contribution!!!

PR #1830, opened 2026-09-03
Add options for import file extensions and type-only imports

### LINKED ISSUE #1829: Add options for import file extensions (.js/.ts) and type-only imports
## Problem

The generated client uses **extensionless, value-only relative imports**:

```ts
import { ApiResponse, Pet, User } from "./data-contracts";
import { ContentType, HttpClient, RequestParams } from "./http-client";
```

This form does not work in two increasingly common TypeScript setups, and there is currently no built-in option to change it.

### 1. Missing file extensions on relative imports

With `moduleResolution: "node16" | "nodenext"`, TypeScript/Node require explicit extensions on relative imports. The generated `./data-contracts` / `./http-client` specifiers fail to resolve at runtime (ESM) and are flagged by the compiler. Projects using `allowImportingTsExtensions` have the same need, but with `.ts` instead of `.js`.

There is no way to tell the generator to emit `./data-contracts.js` (or `.ts`).

### 2. Value imports break `verbatimModuleSyntax` / `isolatedModules`

Type-only names (data contracts, `RequestParams`, `HttpResponse`, and `ContentType` in `enumStyle: "union"`) are imported as **values**. Under `verbatimModuleSyntax` (and `isolatedModules` with certain bundlers) this is an error — those imports must be `import type` (or carry an inline `type` modifier).

Note the http-client import is **mixed**: `HttpClient` is a runtime class (extended/instantiated) and must stay a value import, while `RequestParams`/`HttpResponse` are type-only. So a whole-block `import type` is wrong there — it needs per-specifier inline `type`:

```ts
import { HttpClient, type RequestParams, type HttpResponse } from "./http-client";
```

The data-contracts import, by contrast, is entirely type-only and can use a whole-block `import type`.

Upgrade220:
Closes #1829

## Problem

The generated client emits **extensionless, value-only relative imports**:

```ts
import { ApiResponse, Pet, User } from "./data-contracts";
import { ContentType, HttpClient, RequestParams } from "./http-client";
```

This breaks two increasingly common TypeScript setups, with no built-in option to change it — users must post-process the output by hand:

1. **Missing file extensions.** `moduleResolution: "node16" | "nodenext"` requires explicit extensions on relative imports; the generated `./data-contracts` / `./http-client` fail to resolve at runtime (ESM) and are flagged by the compiler. `allowImportingTsExtensions` needs `.ts` instead.
2. **Value imports break `verbatimModuleSyntax` / `isolatedModules`.** Type-only names (data contracts, `RequestParams`, `HttpResponse`, and `ContentType` in `enumStyle: "union"`) are imported as values, which is an error under those flags.

The http-client import is **mixed**: `HttpClient` is a runtime class and must stay a value import, while `RequestParams`/`HttpResponse` are type-only — so it needs per-specifier inline `type`, not a whole-block `import type`.

## Solution

Two new options, both defaulting to current behavior (no change to existing output):

- **`importFileExtension: "" | ".js" | ".ts"`** (default `""`) — appended to generated relative imports. `.js` for `node16`/`nodenext`, `.ts` for `allowImportingTsExtensions`.
- **`typeOnlyImports: boolean`** (default `false`) — emits `import type` for the wholly type-only data-contracts import, and inline `type` on the mixed http-client import, keeping `HttpClient` a value import. `ContentType` is marked `type` only for `enumStyle: "union"` (where it is a pure type); for `enum`/`const`/`const-enum` it stays a runtime value.

Both are exposed as CLI flags (`--import-file-extension`, `--type-only-imports`).

Implemented at the template level (`templates/modular/api.ejs`, `templates/modular/route-types.ejs`, `templates/default/route-types.ejs`) rather than as a post-processing pass, since the templates already know the import paths and which specifiers are types.

Example — replacing a typical hand-rolled Node16 post-process:

```ts
generateApi({
  importFileExtension: ".js",
  typeOnlyImports: true,
});
```

Produces:

```ts
import type { ApiResponse, Pet, User } from "./data-contracts.js";
import { ContentType, HttpClient, type RequestParams } from "./http-client.js";
```

## Verification

- New spec suite `tests/spec/import-file-extension/` (8 tests): `.js` / `.ts` extensions, default (no extension), whole-block `import type` for data-contracts, inline `type` on the mixed http-client import with `HttpClient` kept as a value, `ContentType` value-vs-type per enum style, and the combined case.
- All existing snapshot tests pass unchanged, confirming default output is byte-identical.
- `bun run build` passes; no new lint warnings/errors introduced.

## Notes

- Adds a changeset (`minor`).
- One unrelated commit adds `/.idea/` and `/.serena/` to `.gitignore`.

---
## Summary by cubic
Adds `importFileExtension` and `typeOnlyImports` options so generated relative imports work with `moduleResolution: node16`/`nodenext`, `allowImportingTsExtensions`, and `verbatimModuleSyntax`/`isolatedModules`. Both default to current behavior, so existing output is byte-identical.

**New Features**
- `importFileExtension` (`""` | `".js"` | `".ts"`) appends an extension to generated relative imports; use `.js` for `node16`/`nodenext` and `.ts` for `allowImportingTsExtensions`.
- Invalid `importFileExtension` values are rejected at runtime; explicit `undefined` normalizes to `""`.
- `typeOnlyImports` emits whole-block `import type` for data-contracts and inline `type` on the mixed http-client import, keeping `HttpClient` as a value import.
- `ContentType` is only marked `type` for `enumStyle: "union"`, where it is a pure type.
- Both options are exposed as CLI flags: `--import-file-extension` and `--type-only-imports`.
- Adds a spec suite covering extension, type-only, combined, and route-type imports.
- Also adds `/.idea/` and `/.serena/` to `.gitignore`.

<sup>Written for commit 5143ae3ad6b998c5628bfe98988c9cac0e5d3e25. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/acacode/swagger-typescript-api/pull/1830?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/import-file-extension-and-type-only-imports.md
+++ b/.changeset/import-file-extension-and-type-only-imports.md
@@ -1,0 +1,14 @@
+---
+"swagger-typescript-api": minor
+---
+
+Add `importFileExtension` and `typeOnlyImports` options
+
+`importFileExtension` (`""` | `".js"` | `".ts"`) appends a file extension to
+generated relative imports, for projects using `moduleResolution: node16`/`nodenext`
+(`.js`) or `allowImportingTsExtensions` (`.ts`).
+
+`typeOnlyImports` emits `import type` for type-only imports (and inline `type` on
+mixed imports such as the http-client import, where `HttpClient` stays a value
+import) for projects using `verbatimModuleSyntax` / `isolatedModules`. `ContentType`
+is only marked `type` for `enumStyle: "union"`, where it is a pure type.
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 /docs/
 /node_modules/
 /tmp-issue-463-run/
+/.idea/
+/.serena/
--- a/index.ts
+++ b/index.ts
@@ -199,6 +199,18 @@ const generateCommand = defineCommand({
         'enum output style: "enum" (default), "union" (T1 | T2 | TN), "const" (as const object + type alias), or "const-enum" (const enum)',
       default: codeGenBaseConfig.enumStyle,
     },
+    "import-file-extension": {
+      type: "string",
+      description:
+        'extension appended to generated relative imports: "" (default), ".js" (moduleResolution node16/nodenext), or ".ts" (allowImportingTsExtensions)',
+      default: codeGenBaseConfig.importFileExtension,
+    },
+    "type-only-imports": {
+      type: "boolean",
+      description:
+        "emit `import type` / inline `type` for type-only imports (verbatimModuleSyntax / isolatedModules)",
+      default: codeGenBaseConfig.typeOnlyImports,
+    },
     "http-client": {
       type: "string",
       description: `http client type (possible values: ${Object.values(
@@ -349,6 +361,12 @@ const generateCommand = defineCommand({
         | "const"
         | "const-enum"
         | undefined,
+      importFileExtension: args["import-file-extension"] as
+        | ""
+        | ".js"
+        | ".ts"
+        | undefined,
+      typeOnlyImports: args["type-only-imports"],
       httpClientType:
         args["http-client"] || args.axios
           ? HTTP_CLIENT.AXIOS
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -61,6 +61,10 @@ export class CodeGenConfig {
   enumStyle: "enum" | "union" | "const" | "const-enum" = "enum";
   /** @deprecated Use enumStyle: "union" instead */
   generateUnionEnums = false;
+  /** CLI flag. Extension appended to generated relative imports: "" (default), ".js", or ".ts". */
+  importFileExtension: "" | ".js" | ".ts" = "";
+  /** CLI flag. Emit `import type` / inline `type` for type-only imports. */
+  typeOnlyImports = false;
   /** CLI flag */
   addReadonly = false;
   enumNamesAsValues = false;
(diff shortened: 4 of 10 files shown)


Upgrade220:
Thanks for the review — both addressed in 4f0dcb2:

1. **CLI validation (index.ts:364):** `importFileExtension` is now validated at runtime in `CodeGenConfig.update`, so any entry point (CLI, config file, JS library) that bypasses the type gets a clear error (`Invalid `importFileExtension` value ".mjs". Expected "", ".js", or ".ts".`) instead of generating imports that don't resolve. Explicit `undefined` normalizes to `""`.
2. **Route-type test coverage:** added tests that generate with `generateRouteTypes: true` and assert the route-types file (`ApiRoute.ts`) gets both the file extension and the `import type` treatment, so `route-types.ejs` is no longer only exercised implicitly.

js2me:
@Upgrade220 thank you!

Upgrade220:
@js2me thank you for maintaining this project!

When to expect release?

js2me:
> @js2me thank you for maintaining this project!
> 
> When to expect release?

Greetings! Release was yesterday

